PR #18, opened 2015-07-13
make advertise to retry on all exceptions for now. Keep it alive.

junchaowu:
@blampe @breerly @abhinav 

Currently advertise only retry for timeout error and protocol error. But when autobahn restarts, tchannel will stop advertising.  Create a large range of exception for retry for now to keep tchannel app alive. This can avoid keep updating current on-boarding service.

That's also what Node does.

### DIFF
--- a/tchannel/tornado/hyperbahn.py
+++ b/tchannel/tornado/hyperbahn.py
@@ -29,8 +29,6 @@ import tornado.gen
 import tornado.ioloop
 
 from ..errors import AdvertiseError
-from ..errors import ProtocolError
-from ..errors import TimeoutError
 from .response import StatusCode
 
 DEFAULT_EXPO_BASE = 1.4  # try this first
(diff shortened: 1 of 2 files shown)


blampe:
Cool. Changelog :)

breerly:
I think this is probably a patch release. @abhinav you agree?

tchannel/tornado/hyperbahn.py
```diff
@@ -68,7 +66,7 @@ def _advertise(tchannel, service):
             headers={'as': 'json'},
             attempt_times=1,
         )
-    except (ProtocolError, TimeoutError) as e:
+    except Exception as e:  # Big scope to keep it alive.
```
breerly: Are you sure we should retry on all exceptions? Seems like we should just keep expanding this list as we find additional exceptions.

PR #20, opened 2015-07-14
Advertise bug

junchaowu:
fix the misuse of periodiccallback and fix time.time() error.
@blampe @abhinav @breerly

### DIFF
--- a/tchannel/tornado/hyperbahn.py
+++ b/tchannel/tornado/hyperbahn.py
@@ -84,7 +84,7 @@ def _advertise_with_backoff(tchannel, service, timeout=None):
     start = time.time()
 
     while True:
-        if timeout and time.time() - start > 30000:
+        if timeout and time.time() - start > 30:
             raise AdvertiseError("Failed to register with Hyperbahn.")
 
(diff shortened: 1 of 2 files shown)


tchannel/tornado/hyperbahn.py
```diff
@@ -84,7 +84,7 @@ def _advertise_with_backoff(tchannel, service, timeout=None):
     start = time.time()
 
     while True:
-        if timeout and time.time() - start > 30000:
+        if timeout and time.time() - start > 30:
```
blampe: I bet we could have used Tornado's `with_timeout` here, oh well.

blampe:
:+1:

tchannel/tornado/hyperbahn.py
junchaowu: I am not sure we can merge 'with_timeout' and periodiccallback together

abhinav: PeriodicCallback isn't special in any way. It just calls the given function repeatedly. The function is responsible for setting timeouts on the asynchronous operations it performs.

PR #23, opened 2015-07-14
rename docs to 'TChannel for Python'

breerly:
Other-words when you click through the main docs to the Python docs you just see "TChannel" then "TChannel".


blampe:
Thanks :)

Issue #30, opened 2015-07-16
clean up API documentation


blampe:
http://tchannel.readthedocs.org/projects/tchannel-python/en/latest/api.html

PR #31, opened 2015-07-16
re-organize API docs and add sync client

blampe:
@breerly @abhinav @junchaowu 

API docs are really bad right now :( #30

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 0.10.4 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Documentation.
 
 
 0.10.3 (2015-07-13)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 0.10.4 (unreleased)
 -------------------
 
-- Documentation.
+- Nothing changed yet.
 
 
 0.10.3 (2015-07-13)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,27 +5,16 @@ API Documentation
 TChannel
 --------
 
-.. automodule:: tchannel.tornado.tchannel
+.. autoclass:: tchannel.tornado.TChannel
     :members:
 
-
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Changelog
 0.10.4 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Documentation.
```
abhinav: Don't need changelog for this. We don't need to cut a release for documentation to be updated, and in this case we haven't even touched the docstrings.

docs/api.rst
```diff
@@ -2,43 +2,55 @@ API Documentation
...
-.. automodule:: tchannel.messages
+.. automodule:: tchannel.tornado.tchannel
     :members:
-    :imported-members:
 
 
-Framing
+Request Dispatch
+----------------
+.. automodule:: tchannel.tornado.dispatch
```
abhinav: This isn't supposed to be used directly anymore.

docs/api.rst
```diff
@@ -2,43 +2,55 @@ API Documentation
 =================
 
 
-Messages
+TChannel
 --------
 
-.. automodule:: tchannel.messages
+.. automodule:: tchannel.tornado.tchannel
```
abhinav: We expect users to import just `tchannel.tornado` here. That re-exports pars of the public API.

docs/api.rst
```diff
@@ -2,43 +2,55 @@ API Documentation
...
-Request
--------
+Thrift
+------
 
-.. automodule:: tchannel.tornado.request
+.. automodule:: tchannel.thrift.client
+    :members:
+
+.. automodule:: tchannel.thrift.server
```
abhinav: Just the client should be fine. Functionality from `tchannel.thrift.server` is made available via the standard TChannel class.

Also, we should move this to `tchannel.tornado.thrift` at some point. We'll have to do it in a backwards compatible way (by keeping the old module around and re-exporting the new one from it).

PR #32, opened 2015-07-16
add a build target for docs

blampe:
Also turns on strict sphinx warnings so we don't accidentally move a module and forget to update docs.

@breerly @abhinav @junchaowu


abhinav:
:+1:

PR #38, opened 2015-07-17
check in keyvalue example

blampe:
Renamed the `set` method to `setValue` to emphasize camel case (and `set` is a syntax error anyway).

I'm not a huge fan of checking in the generated code but it works out of the box.

@abhinav @junchaowu @breerly


examples/keyvalue/keyvalue/client.py
```diff
@@ -0,0 +1,46 @@
...
+from service import KeyValue
+
+
+@gen.coroutine
+def run():
+    app_name = 'keyvalue-client'
+
+    app = TChannel(app_name)
+
+    client_class = client_for('keyvalue-server', KeyValue)
```
abhinav: Make top-level as `KeyValueClient`

examples/keyvalue/keyvalue/client.py
```diff
@@ -0,0 +1,46 @@
...
+
+    client = client_class(app, 'localhost:8889')
```
abhinav: Specifying the arg name might be nicer.

    hostport='localhost:8889'

abhinav:
:+1: 

Minor comments

junchaowu:
lgtm

PR #41, opened 2015-07-18
Fix sync client using 0.0.0.0.0:0 and breaking advertise 

breerly:
Hyperbahn rejects all ephemeral hosts, `0.0.0.0.0`. We need to instead discover the IP.


blampe:
:+1:

PR #43, opened 2015-07-18
TChannel.is_listening(), automatic listen on advertise, do not allow listen twice.

breerly:
cc @blampe @junchaowu @abhinav

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,14 +29,3 @@ script:
 after_success:
   - coveralls -v
 
-#
-# cache dependencies and directories
-# note this can be used on public repos when using container-based infra
-# @see http://docs.travis-ci.com/user/caching/
-# @see http://stackoverflow.com/questions/22024789/speeding-up-travis-ci-dependencies-installation-for-angularjs-project
-# @see https://github.com/rgalanakis/goless/blob/master/.travis.yml
-#
-cache:
-  directories:
-    - .tox
-    - $HOME/.cache/pip
--- a/tchannel/sync/client.py
+++ b/tchannel/sync/client.py
@@ -94,9 +94,6 @@ class TChannelSyncClient(object):
         :returns: first advertise result.
         :raises AdvertiseError: when unable to begin advertising.
         """
-        # begin listening if hasn't already
-        if not self.async_client.is_listening():
-            self.async_client.listen()
 
         @gen.coroutine
         def make_request():
--- a/tchannel/sync/client.py
+++ b/tchannel/sync/client.py
@@ -94,6 +94,9 @@ class TChannelSyncClient(object):
         :returns: first advertise result.
         :raises AdvertiseError: when unable to begin advertising.
         """
+        # begin listening if hasn't already
+        if not self.async_client.is_listening():
+            self.async_client.listen()
 
         @gen.coroutine
         def make_request():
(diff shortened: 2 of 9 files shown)


tchannel/tornado/tchannel.py
```diff
@@ -221,7 +229,16 @@ def listen(self, port=None):
 
         :returns:
             Returns immediately.
+
+        :raises AlreadyListeningError:
+            If listen was already called.
         """
+
+        if self.is_listening():
+            raise AlreadyListeningError(
```
junchaowu: Just throw TChannelError with error message 'already listening'

breerly: No way, I want the client to see the specific exception.

tchannel/sync/client.py
```diff
@@ -94,6 +94,9 @@ def advertise(self, routers, name=None, timeout=None):
         :returns: first advertise result.
         :raises AdvertiseError: when unable to begin advertising.
         """
+        # begin listening if hasn't already
+        if not self.async_client.is_listening():
+            self.async_client.listen()
```
abhinav: Is this necessary? Won't the async client do it automatically?

breerly: Thanks for catching... I started it here and then moved it to the async client; guess I forgot to remove it from here.

tchannel/tornado/tchannel.py
```diff
@@ -123,6 +124,9 @@ def __init__(self, name, hostport=None, process_name=None,
             for peer_hostport in known_peers:
                 self.peers.add(peer_hostport)
 
+        # server created from calling listening
```
breerly: type

breerly:
@abhinav @blampe I think travis's cache is causing threadloop 0.3.0 to be installed... guessing the check in setup.py, `threadloop>=0.3,<0.4` is being ran against the cache, and `0.3.0` passes so it doesn't bother to go out into `pip` and get `0.3.3`. 

I can either update `setup.py` to be `threadloop>=0.3.3,<0.3` or remove the cache. Which is better / more realitic?

breerly:
@abhinav @blampe hmmm... it actually didn't work even when I tightened up the requirements in setup.py... only when I removed the cache... I'm guessing if the cache causes this type of issue then we want to remove it.

tchannel/tornado/tchannel.py
junchaowu: It is weird to call AlreadyListeningError. It could be ServerError, or ServerInitError.

.travis.yml
```diff
@@ -29,14 +29,3 @@ script:
...
   - coveralls -v
 
-#
-# cache dependencies and directories
-# note this can be used on public repos when using container-based infra
-# @see http://docs.travis-ci.com/user/caching/
-# @see http://stackoverflow.com/questions/22024789/speeding-up-travis-ci-dependencies-installation-for-angularjs-project
-# @see https://github.com/rgalanakis/goless/blob/master/.travis.yml
-#
-cache:
```
blampe: How much slower are our runs without this? If it's like... much slower... then we could always clear caches manually in travis if we see this again.

tchannel/tornado/tchannel.py
blampe: Agree with @breerly, `ServerError` is way too generic.

blampe:
Looks like tests are still about a minute so :+1:

abhinav:
LGTM. :shipit:

breerly:
This new Github css looks great

.travis.yml
breerly: Looks like not too much slower

PR #45, opened 2015-07-20
add testing around example code

blampe:
Example code should never break. This also helps make breaking changes more obvious.

@breerly @junchaowu @abhinav


examples/json_client.py
```diff
@@ -22,32 +22,37 @@
...
+    request = tchannel.request(
+        hostport='%s:%s' % (args.host, args.port),
+    )
+
+    response = yield ArgSchemeBroker(JsonArgScheme()).send(
+        request,
+        'hi-json',
+        None,
+        None,
+    )
```
abhinav: :(

abhinav:
:+1:

PR #53, opened 2015-07-22
Initial VCR support

abhinav:
While this works as expected, I don't consider it ready to release yet.  We may want to keep it in the branch for now.

Couple points:

- I'm not entirely sure if patching `choose()` is the best idea. Thoughts?
- For the sake of readable YAML files, I intend to have the system parse Thrift and JSON request and response payloads, and store the parsed objects in the YAML files.
- I also intend to implement the `record_mode`s functionality from vcrpy. It ensures that you don't accidentally make a real request because you broke your test somehow.

Also, note that I've attempted to keep it entirely isolated from the tchannel library. All functionality is under the `tchannel.testing.vcr` module. If we decide to move this to its own library, we can do that fairly easily.


breerly:
+1 on all the points you make in the description.

Let me get into the diff before giving a :ship: :it:

tchannel/testing/vcr/cassette.py
```diff
@@ -0,0 +1,101 @@
+from __future__ import absolute_import
+
+from itertools import chain
+from collections import deque
+
+from . import types
+from . import yaml
+from .exceptions import RequestNotFoundError, UnsupportedVersionError
```
junchaowu: One import per line.

tchannel/testing/vcr/patch.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import absolute_import
+
+from functools import wraps
+
+import mock
+from contextlib2 import ExitStack, contextmanager
```
junchaowu: same here. One import per line.

tchannel/testing/vcr/exceptions.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import absolute_import
+
+
+class VCRException(Exception):
```
junchaowu: @breerly Grayson suggested we call 'Error' instead of 'Exception' long time ago. The base VCRException should also be changed to VCRError

tchannel/testing/vcr/server.py
```diff
@@ -0,0 +1,81 @@
...
+            request.service,
+            arg_scheme=req.arg_scheme,
+            hostport=self.real_peer.hostport,
+            retry='n',
+            # TODO other parameters
+        ).send(
+            request.endpoint,
+            request.headers,
+            request.body,
+            headers=req.headers,  # protocol headers
```
junchaowu: At somepoints, we need to clean up our repo to use standard names. like headers, and protocol_headers. They will become confusing.

tchannel/testing/vcr/cassette.py
```diff
@@ -0,0 +1,101 @@
...
+        :param path:
+            File path at which this cassette will be stored.
+        """
+        self.path = path
+
+        # Number of times interactions from this cassette have been played.
+        self.play_count = 0
+
+        self._available = deque()
+        self._played = deque()
```
junchaowu: what do we do with the '_played'?

junchaowu: It's been created, added,then clear but never used.

tchannel/testing/vcr/cassette.py
```diff
@@ -0,0 +1,101 @@
...
+    def data(self):
+        """Get all known data for this cassette."""
+
+        return deque(
+            chain(self._played, self._available, self._recorded)
+        )
+
+    def _load(self):
+        try:
+            with open(self.path, 'r') as f:
```
junchaowu: It will be better to add a command in Makefile to delete vcr data.

tchannel/testing/vcr/server.py
```diff
@@ -0,0 +1,81 @@
...
+
+class FakeServer(object):
+    """A fake TChannel server to record or replay requests."""
+
+    def __init__(self, cassette, real_peer):
+        self.cassette = cassette
+        self.real_peer = real_peer
+        self.tchannel = TChannel('fake-server')
+
+        self.tchannel.register(self.tchannel.FALLBACK)(self.handle_request)
```
junchaowu: self.tchannel.register(self.tchannel.FALLBACK, handler=self.handle_request) . The way you are using is to use register to return a decorator. Not quite human readable in this way.

tchannel/testing/vcr/exceptions.py
abhinav: Fixed.

tchannel/testing/vcr/cassette.py
abhinav: Huh? File deletion has nothing to do with this code. The `try-catch` is there because you're either creating a new cassette or using an existing one.

tchannel/testing/vcr/server.py
abhinav: Changed

tchannel/testing/vcr/cassette.py
abhinav: Check again. It's used. This information is necessary so that we don't forget already recorded interactions when we're recording new ones for an existing cassette.

tchannel/testing/vcr/server.py
abhinav: I don't see your point? In this specific case, there's no room for adjusting names. It's just using an existing API with the names currently used by that API. The comment is there to clarify that our API currently uses `headers` in `call` to refer to protocol headers. This can't be changed without the refactors we are already planning.

tchannel/testing/vcr/cassette.py
abhinav: Sure. Fixed in all green code.

tchannel/testing/vcr/cassette.py
junchaowu: It's not the code in the python file. It's the Makefile. We should have a command to clean this var data file. And put it in gitignore as well

tchannel/testing/vcr/server.py
junchaowu: I agree. I am not saying changes for this patch.

tchannel/testing/vcr/cassette.py
junchaowu: I know you would put played one into the _played queue. But I don't see _played be used.

abhinav: `data` uses it, and that gets used when actually saving.

tchannel/testing/vcr/cassette.py
abhinav: No. Why would we do that? That completely defeats the point of using a vcr library. You *want* to check the YAML files into your source repo along with your tests so that when you re-run the tests, your tests don't end up making another actual request to the dependent service.

breerly:
Direction looks good - :shipit: to `vcr` branch

tchannel/testing/vcr/cassette.py
junchaowu: It looks strange to me. data will take all three queues of 'interactions' and put them into file. But we it loads up them. These interactions are not classified as played, available, recorded any more. That's why I wonder if this _played is necessary. Because whey you load data from file, actually all are available ones, _played can actually be the _available.

tchannel/testing/vcr/cassette.py
junchaowu: That's true. Any way that's the generated file. For example if you change the test file and make a change to our request, response objects as well, They old data in the file may no longer useful for the new request, response. If at the moment, you run the test, test will broken because the vcr will try to load from old yaml file. That's why we should have a command similar as 'make clean' in the Makefile so that we can clean the old vcr file.

tchannel/testing/vcr/cassette.py
abhinav: Whether an interaction is played or not is knowledge limited/known only for one use of the cassette. When you load a cassette, they're all available in that they can all be replayed. As you play some of them, they're moved to the `played` deque because they're no longer available to be played -- you don't want to allow duplicate plays of the same interaction. During that session, if you record more interactions, they're added to the `recorded` deque because they're also not available to be replayed -- you don't want to prevent someone from making a request twice if they need to, especially if they're testing some retry logic. When you save the cassette, you want all these interactions to be saved--those that were played, those that were not played, and those that were recorded. After that, everything is available because that session with the cassette is done.

tchannel/testing/vcr/cassette.py
abhinav: If you're in the process of still writing your test, you shouldn't be using VCR. Once you've written your test, you add VCR on it so that you're no longer making actual requests to the service. If at some point you change the behavior in the test so that the recording is no longer valid, it's on you to modify/delete the recording. A `make clean` command that deletes your test data doesn't make much sense IMO.

Also, some of your concerns around workflow may be alleviated by the Record Modes feature. See #60. Also have a look at vcrpy's documentation when you get a chance.

abhinav:
I'm gonna close this because this review is technically done. We're going to merge this into master when it's ready.

PR #54, opened 2015-07-23
add thrift transport header

junchaowu:
add thrift transport header & fix zipkin trace
@breerly @abhinav

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,8 +5,8 @@ Changelog
 0.12.1 (unreleased)
 -------------------
 
-- Add transport header support for thrift call.
-- Fix zipkin trace.
+- Add support for specifying transport headers for Thrift clients.
+- Always pass ``shardKey`` for TCollector tracing calls. This fixes Zipkin tracing for Thrift clients.
 
 
 0.12.0 (2015-07-20)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,8 @@ Changelog
 0.12.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Add transport header support for thrift call.
+- Fix zipkin trace.
 
 
 0.12.0 (2015-07-20)
--- a/tchannel/thrift/client.py
+++ b/tchannel/thrift/client.py
@@ -32,7 +32,10 @@ from .scheme import ThriftArgScheme
 from .util import get_service_methods
 
 # Generated clients will use this base class.
-_ClientBase = namedtuple('_ClientBase', 'tchannel hostport service trace')
+_ClientBase = namedtuple(
+    '_ClientBase',
+    'tchannel hostport service trace protocol_headers'
+)
 
 
 def client_for(service, service_module, thrift_service_name=None):
@@ -75,15 +78,21 @@ def client_for(service, service_module, thrift_service_name=None):
 
     method_names = get_service_methods(service_module.Iface)
 
-    def new(cls, tchannel, hostport=None, trace=False):
+    def new(cls, tchannel, hostport=None, trace=False, protocol_headers=None):
         """
         :param tchannel:
             TChannel through which the requests will be sent.
         :param hostport:
             Address of the machine to which the requests will be sent, or None
             if the TChannel will do peer selection on a per-request basis.
+        :param trace:
+            Flag to turn on/off trace in tchannel.
+        :param protocol_headers:
+            Protocol headers of tchannel.
         """
-        return _ClientBase.__new__(cls, tchannel, hostport, service, trace)
+        return _ClientBase.__new__(
+            cls, tchannel, hostport, service, trace, protocol_headers
+        )
 
     new.__name__ = '__new__'
     methods = {'__new__': new}
@@ -152,8 +161,9 @@ def generate_method(service_module, service_name, method_name):
                 hostport=self.hostport, service=self.service
             ),
             endpoint,
-            {},  # TODO: Figure out how to receive headers for the call
+            {},
             call_args,
+            protocol_headers=self.protocol_headers,
             traceflag=self.trace
         )
 
--- a/tchannel/thrift/client.py
+++ b/tchannel/thrift/client.py
@@ -86,9 +86,9 @@ def client_for(service, service_module, thrift_service_name=None):
             Address of the machine to which the requests will be sent, or None
             if the TChannel will do peer selection on a per-request basis.
         :param trace:
-            Flag to turn on/off trace in tchannel.
+            Whether Zipkin tracing is enabled.
         :param protocol_headers:
-            Protocol headers of tchannel.
+            Protocol-level headers to send with the request.
         """
         return _ClientBase.__new__(
             cls, tchannel, hostport, service, trace, protocol_headers
--- a/tchannel/tornado/broker.py
+++ b/tchannel/tornado/broker.py
@@ -64,8 +64,11 @@ class ArgSchemeBroker(object):
         return handler(req, resp, proxy)
 
     @tornado.gen.coroutine
-    def send(self, client, endpoint, header, body, traceflag=None,
-             attempt_times=None, ttl=None,
+    def send(self, client, endpoint, header, body,
+             protocol_headers=None,
+             traceflag=None,
+             attempt_times=None,
+             ttl=None,
              retry_delay=None):
         """Serialize and deserialize header and body into certain format
             based on arg scheme.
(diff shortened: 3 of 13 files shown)


tchannel/tornado/broker.py
```diff
@@ -86,11 +89,13 @@ def send(self, client, endpoint, header, body, traceflag=None,
         except Exception as e:
             raise TChannelError(e.message)
 
+        protocol_headers = protocol_headers or {}
+        protocol_headers['as'] = self.arg_scheme.type()
```
abhinav: Is it possible that someone manually specifies the `as`? If yes, use `setdefault`? Otherwise this is fine.

tchannel/zipkin/tracers.py
```diff
@@ -138,20 +139,24 @@ def __init__(self, tchannel):
         from ..thrift import client_for
 
         # TODO add different thrift service name
-        TCollectorClient = client_for('tcollector', TCollector)
-
+        self.TCollectorClient = client_for('tcollector', TCollector)
```
abhinav: Do this at the top-level? It's a global, unchanging object.

abhinav: Also the thrift service name TODO is not valid.

tchannel/thrift/client.py
```diff
@@ -75,15 +78,21 @@ def post_comment(articleId, msg, hostport=None):
...
         """
         :param tchannel:
             TChannel through which the requests will be sent.
         :param hostport:
             Address of the machine to which the requests will be sent, or None
             if the TChannel will do peer selection on a per-request basis.
+        :param trace:
+            Flag to turn on/off trace in tchannel.
+        :param protocol_headers:
+            Protocol headers of tchannel.
```
abhinav: For `trace`,

    Whether Zipkin tracing is enabled.

For `protocol_headers`

    Protocol-level headers to send with the request.

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -107,3 +125,27 @@ def test_zipkin_trace(trace_server):
...
 
     assert parent_span_id == span_id
+
+
+@pytest.mark.gen_test
+def test_tcollector_submit(monkeypatch, trace_server):
+
+    tchannel = TChannel(name='test')
+
+    hostport = 'localhost:%d' % trace_server.port
```
abhinav: FYI MockServer now provides a hostport attribute. You can just do `trace_server.hostport` instead.

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -107,3 +125,27 @@ def test_zipkin_trace(trace_server):
...
+    tchannel.peers.get(hostport)
```
abhinav: Better to specify it as a known peer when you construct the TChannel.

    TChannel('test', known_peers=[trace_server.hostport])

CHANGES.rst
```diff
@@ -5,7 +5,8 @@ Changelog
 0.12.1 (unreleased)
```
breerly: Let's bump this to `0.13.0` so that new functionality can collect under it.

CHANGES.rst
```diff
@@ -5,7 +5,8 @@ Changelog
...
 -------------------
 
-- Nothing changed yet.
+- Add transport header support for thrift call.
+- Fix zipkin trace.
```
breerly: Can you be more descriptive? What about zipkin trace was broken? Can we not fix this in a patch release without introducing any new functionality?

abhinav: - Add support for specifying transport headers for Thrift clients.

BTW, if I understand correctly, the fix for Zipkin tracing for Thrift clients was basically to start passing the shard key, right? In that case, change the second one to,

    - Always pass ``shardKey``s for TCollector tracing calls. This fixes Zipkin tracing for Thrift clients.

Basically, saying "Fixed X" isn't very informative. It's as good as "It doesn't work" in bug reports.

tchannel/thrift/client.py
```diff
@@ -75,15 +78,21 @@ def post_comment(articleId, msg, hostport=None):
 
     method_names = get_service_methods(service_module.Iface)
 
-    def new(cls, tchannel, hostport=None, trace=False):
+    def new(cls, tchannel, hostport=None, trace=False, protocol_headers=None):
```
breerly: Do we need to do anything to the sync client to support as well?

tchannel/tornado/peer.py
```diff
@@ -624,7 +624,7 @@ def send_with_retry(self, request, peer, attempt_times, retry_delay):
                 )
 
                 if not connection:
-                    raise
+                    raise protocol_error
```
breerly: Looks like this constitutes a breaking change? We'll want to document that in `UPGRADING.rst` and `CHANGES.rst`.

CHANGES.rst
junchaowu: After this patch. I will bump it with separate one.

tchannel/tornado/peer.py
abhinav: not a breaking change. this was previously a re-raise of the same exception, but since we're inside a coroutine, the yield could forget this exceptino

tchannel/zipkin/tracers.py
```diff
@@ -40,13 +40,17 @@
...
+from .formatters import i64_to_base64
 from .thrift import TCollector
 from .thrift import constants
+from ..thrift import client_for
 
 log = logging.getLogger('zipkin_tracing')
 
 zipkin_log = logging.getLogger('zipkin')
 
+TCollectorClient = client_for('tcollector', TCollector)
```
breerly: This introduces very tight Thrift and TCollector couping - are we absolutely sure we want that? cc @abhinav @blampe

tchannel/tornado/peer.py
abhinav: so more like a bugfix

tchannel/thrift/client.py
junchaowu: Yes we do. You need to update it with sync client. But it's not high priority. You can do that after the coming api call change.

tchannel/zipkin/tracers.py
junchaowu: This one is for zipkin tracer and we use thrift to submit zipkin spans.

tchannel/tornado/peer.py
abhinav: To explain that further:

    try:
        foo()
    except Exception:
        # stuff
        raise

Usually works, but if you're using coroutines and yield one inside the Exception block, and the called coroutine returns with `raise Return`, the system's knowledge of "the last exception" changes.

breerly: Can we fix this in a separate PR and cut a patch release `0.12.1`?

tchannel/thrift/client.py
breerly: Seems like we should add it now, so that both async and sync get trace support. What do you think @abhinav @blampe ?

tchannel/zipkin/tracers.py
abhinav: TCollector and Thrift are already tightly coupled right now.

In fact, we currently have a hard-dependency on the thrift library regardless of whether you use Thrift or not. We need to figure that situation out at some point.

tchannel/zipkin/tracers.py
```diff
@@ -135,23 +139,23 @@ def __init__(self, tchannel):
...
+    def record(self, traces):
         for (trace, annotations) in traces:
-            f = self.client.submit(thrift_formatter(trace, annotations))
-            f.add_done_callback(submit_callback)
+            client = TCollectorClient(
+                self._tchannel,
+                protocol_headers={'shardKey': i64_to_base64(trace.trace_id)}
+            )
+            f = client.submit(thrift_formatter(trace, annotations))
+            f.add_done_callback(self.submit_callback)
```
abhinav: So, if I understand correctly, the sole reason submit_callback is a public method is so that you can override it and get the real result for a test. That's not a very good approach to solve this problem.

What if you returned those futures as a list (in addition to adding the `submit_callback` for logging -- which doesn't have too be a public method) from the `record` call? The caller of record can choose to yield those futures and get the results if necessary. (You can yield a list of futures to get a list of the results.)

Generally, if you're making an asynchronous call, it is beneficial if your caller can wait for/get the result.

junchaowu: That's a great idea.

abhinav: This is good, but you want to move submit_callback back. It's okay if it's a lambda or a `_` method. Basically, it's not part of the public interface.

abhinav:
lgtm

tchannel/thrift/client.py
blampe: +1 to zipkin traces from the sync client.

PR #56, opened 2015-07-23
Add instantiation-time support of protocol headers to sync thrift client.

junchaowu:


tchannel/sync/thrift.py
```diff
@@ -70,11 +70,15 @@ def client_for(service, service_module, thrift_service_name=None):
...
-    def init(self, tchannel_sync, hostport=None, trace=False):
+    def init(self, tchannel_sync,
+             hostport=None,
+             trace=False,
+             protocol_headers=None):
         self.async_thrift = self.__async_client_class__(
             tchannel_sync._async_client,
             hostport,
             trace,
+            protocol_headers,
```
abhinav: Can you change this to use keyword args for all arguments?

The number of positional args here is getting unwieldy.

abhinav:
Looks good besides minor styling change.

tchannel/sync/thrift.py
breerly: What are all the args are being overloaded into `protocol_headers`? I just assumed it was a free-form dict...

abhinav:
LGTM. Address @breerly's concerns and :shipit:.

tchannel/sync/thrift.py
junchaowu: Yes. it is a dict.

PR #57, opened 2015-07-23
Address test_examples flakiness.

abhinav:
Also clean up handlers registered with sample servers.

@breerly @junchaowu


breerly:
baller

PR #59, opened 2015-07-23
Address test_examples flakiness

abhinav:


abhinav:
Connection refused after `wait_until_connectable` explicitly confirms that the port is accepting connections? meh. Will investigate later.

PR #61, opened 2015-07-24
Build and install Thrift before running tests.

abhinav:
Not sure if we want to keep this. I wanna see how long it takes since this basically installs from source.


abhinav:
Getting Thrift installed will take more effort. Punting.

Issue #12, opened 2015-07-11
Drop two-phase API in favor of single call method.


abhinav:
The two phase `request(..).send(..)` API seems unnecessary. Let's converge it into a single call.

Assigning to @breerly  because he expressed interest in doing this.

breerly:
Whipped up a gist for the proposed API. Please comment:

https://gist.github.com/breerly/ca4f6b9c17861327aa8a

Grayson:
@abhinav, I believe you've tagged in the wrong Grayson.

abhinav:
Oh, right. Yes, the `@` was wrong. Thanks.

junchaowu:
Oh here it is.

breerly:
fixed in 0.15.0

PR #64, opened 2015-07-24
TChannel.call

breerly:
Will be updating this as I go to encourage early feedback.

- [x] `tchannel.call()`
- [x] `tchannel.raw()`
- [x] `tchannel.json()`
- [ ] `tchannel.thrift()`

Other details/concerns:

- [x] don't shadow built-in types, change `formats` to `schemes`. 
- [ ] test all exception types coming out of call.
- [x] main api's should autocomplete w jedi & pycharm (good as it gets, __call__ not yet supported).
- [ ] switch `header` to `headers` everywhere (even with raw) - what about server impl?
- [x] don't use thrift cli for tests
- [ ] determine if `stream()` funcs will affect api
- [ ] do we need a top level `request`, since we now have a `response`?
- [ ] cover all request transport headers, for example, does timeout even get used?
- [ ] test thrift serialization
- [ ] demo w team

Addresses #12


breerly:
FYI did a bunch of research on other RPC's "request" API here - https://gist.github.com/breerly/ca4f6b9c17861327aa8a

tchannel/tchannel.py
```diff
@@ -0,0 +1,42 @@
...
+        if headers is None:
+            headers = {}
+
+        if timeout is None:
+            timeout = DEFAULT_TIMEOUT
+
+    @gen.coroutine
+    def call_raw(self, service, endpoint, body, headers=None, timeout=None):
+
+        response = yield self.call('raw', service, endpoint, body, timeout)
```
breerly: @abhinav @junchaowu wondering why we don't have the common arg-scheme in glossary... should I add?

tchannel/tornado/tchannel.py
```diff
@@ -212,6 +213,12 @@ def request(self,
 
             Defaults to 'c'.
         """
+        # TODO: do not warn if being called from tchannel.TChannel
```
breerly: will probably warn in init as well

tchannel/tchannel.py
```diff
@@ -0,0 +1,42 @@
...
+            timeout = DEFAULT_TIMEOUT
+
+    @gen.coroutine
+    def call_raw(self, service, endpoint, body, headers=None, timeout=None):
+
+        response = yield self.call('raw', service, endpoint, body, headers, timeout)
+
+        raise gen.Return(response)
+
+    @gen.coroutine
```
junchaowu: I would think it is too early to add coroutine decorator here. If you don't do yield inside the function, and delegate the coroutine work to some internal method. Here just return a future.

breerly: Good point!

breerly: @junchaowu thanks for the early feedback!

tchannel/tchannel.py
```diff
@@ -0,0 +1,41 @@
...
+            formatters = DEFAULT_FORMATS
+
+        # set formatters
+        for f in formatters:
+            # if not abc, blow up
+            if not issubclass(f, Formatter):
+                raise Exception("not valid formatter")
+
+            # init and set on self
+            f = f(self)
```
breerly: Notice how each Formatter gets the tchannel instance

tchannel/tchannel.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # if no formatters, use default
+        if not formatters:
+            formatters = DEFAULT_FORMATS
+
+        # set formatters
+        for f in formatters:
+            # if not abc, blow up
+            if not issubclass(f, Formatter):
+                raise Exception("not valid formatter")
```
breerly: I'll obviously define a more specific exception

tchannel/format.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import (
```
blampe: formats/base.py?

tchannel/tchannel.py
```diff
@@ -0,0 +1,41 @@
...
+
+            # init and set on self
+            f = f(self)
+            setattr(self, f.name, f)
```
breerly: After this point you'd have:

```
tchannel = TChannel()
tchannel.raw.call
tchannel.raw.stream
tchannel.json.call
tchannel.json.stream
tchannel.thrift.call
tchannel.thrift.stream
```

breerly: Also, if they passed in custom formatter you could have:

```
tchannel.proto3.call
tchannel.proto3.stream
```

for example

tchannel/format.py
breerly: i want to keep the pluralized module all homogenous types

breerly: so:

```
format - top-level details defining format - so like constants, config, abcs, etc...
formats - implementors of the Formatter abc
```

breerly: Although it's not out of hand to put `formats` in the `format` package

abhinav: I'm in favor of putting it inside the `formats` package. The `formats` package would basically expose standard formats supported by TChannel and the machinery to integrate your own custom formats.

abhinav: And I mean the most common use case doesn't involve importing any of these modules directly anyway.

tchannel/tornado/tchannel.py
abhinav: The initial planned implementation of `call` will be calling this method. You don't want to tag it as deprecated until the flow is reversed -- the old implementation is calling the new implementation.

tchannel/tchannel.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, formatters=None):
+
+        # if no formatters, use default
+        if not formatters:
+            formatters = DEFAULT_FORMATS
+
+        # set formatters
+        for f in formatters:
+            # if not abc, blow up
+            if not issubclass(f, Formatter):
```
abhinav: I don't know if we need to check `isinstance`/`issubclass`. Duck typing. Just accept whatever and assume it has the required methods.

abhinav:
Looks mostly good to me. Couple nits.

Also, once we have this working, let's move on to the descriptor approach.

tchannel/format.py
breerly: Noted. It might come to past before we merge this that we don't need the abc.

tchannel/tchannel.py
breerly: yeah, that's what I was saying about potentially not needing the abc

tchannel/tornado/tchannel.py
breerly: @abhinav I was going to detect if it was being called from `tchannel.TChannel`... is that too nasty?

abhinav: yes. please don't.

breerly: @abhinav :P - I don't feel strongly either way, was just sketchin' out what I think we want. At this stage, I'm looking for ideas on how to accomplish deprecating these properly instead of just "dont do this".

abhinav: I was thinking: We make `call` implemented in terms of `request.send` available, release that, then we invert refactor and invert it so that `request.send` is implemented in terms of `call`, and that's when we add the deprecated warning to `call`. We release that. Then we release a new version that doesn't have `request.send`.

breerly: Good idea - let's roll this way unless there are objections.

tchannel/tchannel.py
```diff
@@ -0,0 +1,75 @@
...
+        # TODO - dont use asserts for public API
+        assert format, "format is required"
+        assert service, "service is required"
+        assert arg1, "arg1 is required"
+
+        # default args
+        if arg2 is None:
+            arg2 = ""
+        if arg3 is None:
+            arg3 = ""
```
breerly: seems like the `arg2` & `arg3` params should have `None` as the default in the method signature...

tests/formats/test_json.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.mark.gen_test
+@pytest.mark.call
+def test_call_should_get_response(mock_server):
+
+    endpoint = 'endpoint'
+    body = {
+        'key': 'value'
+    }
+
+    mock_server.expect_call(endpoint, scheme='json').and_write(
```
breerly: todo - change expect_call param to use `format` instead of `scheme`.

tchannel/__init__.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+from .tchannel import TChannel  # noqa
+from .response import Response  # noqa
```
breerly: Not sure if this needs to be top level actually...

breerly:
Reminder to not forget to test exceptions

breerly:
Also, turns out `format` is a built-in function, so we shouldn't shadow it. This is what's causing PyCharm to warn and vim to highlight it like it's a reserved word. Will change the module name to either `schemes`, `argschemes`, or `arg_schemes`.

tests/test_tchannel.py
```diff
@@ -0,0 +1,67 @@
...
+        assert isinstance(scheme, f)
+
+
+@pytest.mark.gen_test
+@pytest.mark.call
+def test_call_should_get_response():
+
+    # Given this test server:
+
+    server = DeprecatedTChannel(name='server')
```
blampe: Why use `DeprecatedTChannel` instead of `TChannel`?

tests/formats/test_raw.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+import pytest
+import tornado
+
+from tchannel import TChannel, schemes
```
blampe: Renamed tests/formats -> tests/schemes.

tchannel/__init__.py
blampe: Nuke?

tchannel/schemes/json.py
```diff
@@ -0,0 +1,53 @@
...
+    """Semantic params and serialization for json."""
+
+    NAME = JSON
+
+    def __init__(self, tchannel):
+        self._tchannel = tchannel
+
+    @gen.coroutine
+    def __call__(self, service, endpoint, body=None,
+                 header=None, timeout=None):
```
blampe: headers

tchannel/response.py
```diff
@@ -0,0 +1,19 @@
...
+)
+
+from . import transport
+
+__all__ = ['Response']
+
+
+class Response(object):
+    def __init__(self, header, body, transport):
+        self.header = header
```
blampe: headers

breerly: @blampe this is actually a weird thing because for `raw` it's a free-form "header block", but for `json` & `thrift` it is basically key-value pairs... I suppose `headers` is more right for the majority, even if `raw` it would be free-form. 

@abhinav @junchaowu what do you guys think?

tests/formats/test_raw.py
breerly: aaah... good catch

tests/test_tchannel.py
breerly: I'll switch once I add register to `TChannel` :)

tests/formats/test_json.py
```diff
@@ -0,0 +1,59 @@
...
+    server.listen()
+
+    # Make a call:
+
+    tchannel = TChannel(name='test')
+
+    resp = yield tchannel.json(
+        service=server.hostport,
+        endpoint='endpoint',
+        header={'req': 'header'},
```
breerly: @blampe like here you're right, `headers` makes way more sense, unless we always want people to think of it as the `header` block... but that makes them have to care about the transport, probably more than makes sense...

tchannel/response.py
junchaowu: consistent。 If we use header, then header everywhere. If headers, headers everywhere.

tchannel/schemes/json.py
junchaowu: Prefer to keep the same order of parameters.  first header(s) then body.

tchannel/schemes/raw.py
```diff
@@ -0,0 +1,32 @@
...
+    def __call__(self, service, endpoint, body=None,
+                 header=None, timeout=None):
+
+        return self._tchannel.call(
+            scheme=self.NAME,
+            service=service,
+            arg1=endpoint,
+            arg2=header,
+            arg3=body,
+            timeout=timeout,
```
junchaowu: We shall fill all the arguments of call as we implement them even if they have default value. It can help us avoid ambiguity.

tchannel/transport.py
```diff
@@ -0,0 +1,41 @@
...
+    args = {}
+    args['caller_name'] = data.get(CALLER_NAME)
+    args['claim_at_start'] = data.get(CLAIM_AT_START)
+    args['claim_at_finish'] = data.get(CLAIM_AT_FINISH)
+    args['failure_domain'] = data.get(FAILURE_DOMAIN)
+    args['retry_flags'] = data.get(RETRY_FLAGS)
+    args['scheme'] = data.get(SCHEME)
+    args['shard_key'] = data.get(SHARD_KEY)
+    args['speculative_exe'] = data.get(SPECULATIVE_EXE)
+
```
junchaowu: data.get(xxx) can throw exception if the key doesn't exist.

tchannel/schemes/json.py
breerly: nice, will do :)

tchannel/transport.py
breerly: I initially thought that, but in this case I just want optimistically create the kwarg bag. Basically - if it's there, do our best to convert the wire args to our valid kwargs.

tchannel/response.py
breerly: I'll switch everything to `headers` then.

breerly: Feels a bit weird for raw since plural var's typically indicate iterability

tchannel/schemes/raw.py
breerly: hmmm?

tchannel/thrift/generated.py
```diff
@@ -0,0 +1,95 @@
...
+
+class ThriftRequestMaker(object):
+
+    def __init__(self, service, thrift_module):
+        self.service = service
+        self.thrift_module = thrift_module
+
+    def _make_request(self, method_name, args, kwargs):
+
+        # TODO what to do w args and kwargs?
```
breerly: remove this todo, already done

tchannel/schemes/thrift.py
```diff
@@ -0,0 +1,45 @@
...
+        header = serializer.serialize_header(header)
+        body = serializer.serialize_body(request.call_args)
+
+        response = yield self._tchannel.call(
+            scheme=self.NAME,
+            service=request.service,
+            arg1=request.endpoint,
+            arg2=header,
+            arg3=body,
+            hostport=hostport
```
breerly: Don't forget to remove the distinction between service & hostport if they are never used in practice simultaneously.

tchannel/thrift/generated.py
```diff
@@ -0,0 +1,95 @@
...
+    absolute_import, division, print_function, unicode_literals
+)
+
+import inspect
+import types
+
+from .util import get_service_methods
+
+
+def from_thrift_module(service, thrift_module):
```
breerly: this might be better if api was like:

```
from tchannel import thrift

service = thrift.from_module(thrift_module)
service = thrift.from_idl('some_idl.thrift')
```

tests/schemes/test_json.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.gen_test
+@pytest.mark.call
+def test_call_should_get_response():
+
+    # Given this test server:
+
+    server = DeprecatedTChannel(name='server')
+
+    @server.register('endpoint', schemes.JSON)
```
breerly: wondering if this could just be:

```
@server.json.register('endpoint')
...
```

tchannel/thrift/generated.py
blampe: There's room for confusion between the modules and the schemes though:

```
import tchannel.thrift
vs.
tchannel = TChannel()
response = tchannel.thrfit(...)
```

Personally I think it would be worth re-visiting the library managing the TChannel singleton for you (kind of like how Tornado keeps track of the IOLoop but still exposes enough control for power users). Then you could just interact with modules like `tchannel.thrift` which would use `TChannel` behind the scenes.

tchannel/schemes/json.py
breerly: Thinking about moving headers after body since the more common case is sending body w/o application headers

blampe: +1000 yess

breerly:
@abhinav @junchaowu @blampe - sorry for all the generated code, makes this PR a PITA to understand. I want to demo this in-person tomorrow after my jury duty.

tchannel/thrift/generated.py
breerly: Sure. For now we might be able to just keep going down the path I'm doin' now like so:

```
from tchannel import TChannel, from_thrift_module, from_thrift_idl
```

tchannel/tchannel.py
```diff
@@ -0,0 +1,71 @@
...
+        if timeout is None:
+            timeout = DEFAULT_TIMEOUT
+
+        # get operation
+        # TODO lets treat hostport and service as the same param
+        operation = self._dep_tchannel.request(
+           service=service,
+           hostport=hostport or service,
```
breerly: Another reminder that we shouldn't need `service` and `hostport` at call-sites.

breerly:
Closing since I destroyed this PR with generated code and I want to merge into `call`, not `master`.

PR #67, opened 2015-07-29
Rewrite patching logic to use a Thrift service for request proxying.

abhinav:
The VCR functionality is implemented by a Thrift service called VCRProxy.

Also,

* Instead of patching `PeerGroup.choose`, we now patch `TChannel.request`.
* Added support for matcher customization for cassettes.

@breerly @blampe @junchaowu


abhinav:
FYI I mean to merge the vcr branch into master after this is reviewed. I think it is ready for use. We can address the TODOs and parsing of request/response bodies later.

tchannel/testing/vcr/__init__.py
```diff
@@ -22,8 +22,8 @@
 VCR
```
blampe: Oh boo you put this all in __init__.

blampe:
LGTM!

tchannel/testing/vcr/__init__.py
breerly: @abhinav please move out and instead import into init

abhinav: fixed

PR #68, opened 2015-07-30
define claim/cancel message types

blampe:
We were missing these message types. cc @abhinav @breerly @junchaowu 

> This claim message is sent from worker to to worker. The claimed request is referred to by its full zipkin tracing data, which was chosen by the originator of the first request.

> When a worker B receives a claim message from worker A for a tracing T that B doesn't know about, B will note this for a short time. If B later receives a request for T, B will silently ignore T. This is expected to be the common case, because the forwarding router will add some delay after sending to A and before sending to B.

This seems to indicate that workers are communicating directly with each other, or that they can force requests over hyperbahn to end up at a specific instance. Both of those seem like things we've been actively trying to avoid. @Raynos

### DIFF
--- a/tchannel/messages/__init__.py
+++ b/tchannel/messages/__init__.py
@@ -20,28 +20,32 @@
 
 from __future__ import absolute_import
 
-from .types import Types
-from .error import ErrorMessage, ErrorCode, error_rw
-from .common import Tracing, ChecksumType
 from .call_request import CallRequestMessage, call_req_rw
+from .call_request_continue import call_req_c_rw
 from .call_response import CallResponseMessage, call_res_rw
+from .call_response_continue import call_res_c_rw
+from .cancel import CancelMessage, cancel_rw
+from .claim import ClaimMessage, claim_rw
+from .common import Tracing, ChecksumType
+from .error import ErrorMessage, ErrorCode, error_rw
 from .init_request import InitRequestMessage, init_req_rw
 from .init_response import InitResponseMessage, init_res_rw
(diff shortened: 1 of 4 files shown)


kriskowal:
:ok:

@blampe Claim messages in particular are not intended to be routed through Hyperbahn. They are intended to be routed within a service’s cluster.

abhinav:
LGTM

blampe:
@kriskowal how do workers discover each other, though?

Raynos:
@blampe 

cancel and claim are designed for implementing speculative execution. They are an internal part of tchannel and hyperbahn and not really meant to be used as part of the application developer public interface.

The expected workflow is as follows: 

 - On an hyperbahn exit node, if a call request `se` header is set to `2` then:
 - hyperbahn exit node picks two random instances, A & B
 - Send call request to A with claim at finish === B
 - Wait ttl / 2, if A is not finished then send the same call request to B with claim at start === A
 - If A finishes the request then it sends claim to B, if B starts the request then it sends claim to A.
 - Claim messages are cheap to send if your service is a ringpop and is already fully connected + has open connections
 
If both A & B are slow/broken then the request will timeout.

If A is broken, but B is healthy then the request finished in ttl/2
If A is healthy then the request finishes normally.

> This seems to indicate that workers are communicating directly with each other

That is ringpop. Workers communicate directly because everything is ringpop.

blampe:
@Raynos that's what I didn't want to hear, let's talk in person.

PR #71, opened 2015-07-31
TChannel.call - client interface

breerly:
Initial design, research for this can be found at: https://gist.github.com/breerly/ca4f6b9c17861327aa8a

Will be updating this as I go to encourage early feedback.

- [x] `tchannel.call()`
- [x] `tchannel.raw()`
- [x] `tchannel.json()`
- [x] `tchannel.thrift()`

Other details/concerns:

- [x] don't shadow built-in types, change `formats` to `schemes`. 
- [x] test all exception types coming out of call.
- [x] main api's should autocomplete w jedi & pycharm (good as it gets, __call__ not yet supported).
- [x] switch `header` to `headers` everywhere in client api
- [x] don't use thrift cli for tests
- [x] determine if `stream()` funcs will affect api - punt for this PR
- [x] do we need a top level `request`, since we now have a `response`? will handle w server diff
- [x] cover all request transport headers, for example, does timeout even get used?
- [x] test `tchannel.thrift.serializer` - will add TODO instead
- [x] test timeout & retries... maybe w vcr? else it has to be mock/doubles - will add TODO instead
- [x] demo w team
- [x] fix thrift not working unless headers provided
- [x] fix header not being written in thrift exapmle
- [x] service and hostport should remain seperate concepts
- [x] create example files
- [x] test thrift void use-cases
- [x] implement __repr__ on response class - will add TODO instead
- [x] verify tracing still works

Addresses #12


breerly:
cc @abhinav @junchaowu @blampe created this new PR from #64 since that PR got wasted :P

tchannel/response.py
```diff
@@ -0,0 +1,19 @@
...
+    absolute_import, division, print_function, unicode_literals
+)
+
+from . import transport
+
+__all__ = ['Response']
+
+
+class Response(object):
+    def __init__(self, headers, body, transport):
```
abhinav: `__slots__`

tchannel/schemes/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+DEFAULT_NAMES = (
+    RAW,
+    JSON,
+    THRIFT
+)
+
+from .raw import RawArgScheme  # noqa
+from .json import JsonArgScheme  # noqa
+from .thrift import ThriftArgScheme  # noqa
```
abhinav: Is there a reason to put the imports here instead of the top of the file?

tchannel/schemes/json.py
```diff
@@ -0,0 +1,53 @@
...
+            service=service,
+            arg1=endpoint,
+            arg2=headers,
+            arg3=body,
+            timeout=timeout,
+        )
+
+        # deserialize
+        response.headers = json.loads(response.headers)
+        response.body = json.loads(response.body)
```
abhinav: Can we do serialization/deserialization inside call? As in, we pass in something that knows how to serialize and deserialize the header?

tchannel/schemes/json.py
```diff
@@ -0,0 +1,53 @@
...
+    """Semantic params and serialization for json."""
+
+    NAME = JSON
+
+    def __init__(self, tchannel):
+        self._tchannel = tchannel
+
+    @gen.coroutine
+    def __call__(self, service, endpoint, body=None,
+                 headers=None, timeout=None):
```
abhinav: You'll probably need `**kwargs` here? You need to pass in extra configuration parameters over to `call`

tchannel/schemes/thrift.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # deserialize
+        response.headers = serializer.deserialize_headers(
+            headers=response.headers
+        )
+        body = serializer.deserialize_body(
+            body=response.body,
+            result_type=request.result_type
+        )
+        response.body = body.success
```
abhinav: TODO for handling Thrift exceptions and return types here?

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+    absolute_import, division, print_function, unicode_literals
+)
+
+import inspect
+import types
+
+from .reflection import get_service_methods
+
+
+def from_thrift_module(service, thrift_module):
```
abhinav: Not a fan of this name. What is being built from a thrift module?

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+    maker = ThriftRequestMaker(service, thrift_module)
+
+    # create methods that mirror thrift client
+    # and each return ThriftRequest
+    methods = _create_methods(thrift_module)
+
+    # attach to maker
+    for name, method in methods.iteritems():
+        method = types.MethodType(method, maker, ThriftRequestMaker)
+        setattr(maker, name, method)
```
abhinav: This is doing two separate out-of-context loops for no reason. Can you get rid of `create_methods` but keep `create_method`.

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+
+        args_type = getattr(self.thrift_module, method_name + '_args')
+
+        return args_type
+
+    def _get_result_type(self, method_name):
+
+        result_type = getattr(self.thrift_module, method_name + '_result')
+
+        return result_type
```
abhinav: Okay, this is incredibly nitpicky but can you reduce the number of empty lines here.

    def get_foo():
    
        foo = ...

        return foo

There's too much stuttering here for my taste.

tchannel/schemes/thrift.py
```diff
@@ -0,0 +1,50 @@
...
+
+class ThriftArgScheme(object):
+
+    NAME = THRIFT
+
+    def __init__(self, tchannel):
+        self._tchannel = tchannel
+
+    @gen.coroutine
+    def __call__(self, request=None, headers=None, timeout=None, hostport=None):  # TODO remove hostport here...
```
blampe: Lint:

```
def __call__(
    self,
    request=None,
    headers=None,
    timeout=DEFAULT_TIMEOUT,
    hostport=None,
):
```

tchannel/tchannel.py
```diff
@@ -0,0 +1,71 @@
...
+            name, hostport, process_name, known_peers, trace
+        )
+
+        # set arg schemes
+        self.raw = schemes.RawArgScheme(self)
+        self.json = schemes.JsonArgScheme(self)
+        self.thrift = schemes.ThriftArgScheme(self)
+
+    @gen.coroutine
+    def call(self, scheme, service, arg1, arg2=None, arg3=None, timeout=None, hostport=None):
```
blampe: Lint

```
def call(
    self,
    scheme,
    service,
    arg1,
    arg2=None,
    arg3=None,
    timeout=DEFAULT_TIMEOUT,
    hostport=None,
):
```

tchannel/schemes/json.py
blampe: +1 but we can move the internals around later.

tchannel/sync/thrift.py
```diff
@@ -20,7 +20,7 @@
 
 from __future__ import absolute_import
 
-from tchannel.thrift.util import get_service_methods
+from tchannel.thrift.reflection import get_service_methods
```
blampe: I like the reflection name. Maybe move `client_for` in there?

tchannel/thrift/module.py
breerly: yea ok

tchannel/response.py
breerly: thx for 'ze reminder :)

tchannel/schemes/__init__.py
breerly: yeah - you want to tame import order in a given packages init. In this case, the *Scheme types use the constants defined above them.

tchannel/schemes/json.py
breerly: Yeah - I haven't thought about that yet. I'll grab you in a bit to jam.

tchannel/schemes/json.py
breerly: I think I disagree - `call` is the workhorse func, I don't want it doing or invoking any serialization.

tchannel/schemes/thrift.py
breerly: yeah - I'm ganking hostport now

tchannel/sync/thrift.py
breerly: Was going to leave it in place since it will be removed in not too long

tchannel/tchannel.py
breerly: removing hostport will fix

tchannel/transport.py
```diff
@@ -0,0 +1,41 @@
...
+    args = {}
+    args['caller_name'] = data.get(CALLER_NAME)
+    args['claim_at_start'] = data.get(CLAIM_AT_START)
+    args['claim_at_finish'] = data.get(CLAIM_AT_FINISH)
+    args['failure_domain'] = data.get(FAILURE_DOMAIN)
+    args['retry_flags'] = data.get(RETRY_FLAGS)
+    args['scheme'] = data.get(SCHEME)
+    args['shard_key'] = data.get(SHARD_KEY)
+    args['speculative_exe'] = data.get(SPECULATIVE_EXE)
+
```
junchaowu: We haven't limited whether user can set their key-value into transport header. We probably don't want to lose those user customized header either.

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+
+        call_args = args_type()
+        for name, value in params.items():
+            setattr(call_args, name, value)
+
+        return call_args
+
+
+class ThriftRequest(object):
+
```
abhinav: `__slots__`

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+        method = _create_method(method_name)
+        methods[method_name] = method
+
+    return methods
+
+
+def _create_method(method_name):
+
+    def method(self, *args, **kwargs):
```
abhinav: So one thing worth doing might be to pass in a reference to the `Iface.<method>` method and decorating this with `@functools.wraps(that_function)`. It'll copy over the docstrings, etc. 

If you *really* wanted to pretend to be the `Iface.<method>` function, you might be able to use `wrapt` for that but that's probably not worth it right now.

breerly: @abhinav that would be amazing

tchannel/schemes/json.py
junchaowu: I remember there is an issue with using kwargs. It will be unclear for caller to know those arguments.

tchannel/thrift/module.py
abhinav: Also, I don't think this is the best approach to generating objects dynamically. You should use

    bases = (thrift_module.Iface, ThriftRequestMaker)
    type(service_name + 'RequestBuilder', bases, dct)

Where `dct` is the dictionary from method name to `create_method`-generated methods. You still get the same behavior of passing all calls through `make_request` but you're now generating a separate class for the service instead of dynamic `setattr` like that (and it now also inherits the service Iface class). You could theoretically even use `slots` or named tuple with this approach if you wanted to.

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+        result_type = getattr(self.thrift_module, method_name + '_result')
+
+        return result_type
+
+    def _get_call_args(self, method_name, args, kwargs):
+
+        args_type = self._get_args_type(method_name)
+
+        params = inspect.getcallargs(
+            getattr(self.thrift_module.Iface, method_name),
```
abhinav: You'll `getattr` on the class for all calls if you do this. Resolve this information only once when you're generating the methods.

abhinav: Same for the `get_args_type` and `get_result_type` functions.

abhinav: So, you want some sort of ThriftMethod class that contains all information about the method not bound to specific parameters. And then do `make_request(thrift_method, parameters)` where parameters is the `getcallargs` resolved arguments.

tchannel/thrift/module.py
```diff
@@ -0,0 +1,114 @@
...
+
+    methods = {}
+    method_names = get_service_methods(thrift_module.Iface)
+
+    for method_name in method_names:
+
+        method = _create_method(method_name)
+        methods[method_name] = method
+
+    return methods
```
abhinav: I would get rid of this method. It doesn't save you anything. Instead of doing this in the caller,

    methods = create_methods(...)
    for name, method in methods.items():
        # ...

Do,

    for name in get_service_methods(...):
        method = _create_method(...)
        # ...

tchannel/schemes/json.py
abhinav: call is the workhorse function. you want it doing pre and post processing too. It doesn't need to how that serialization/deserialization is happening. It just gets an object that does. The default implementation for that object could just be the identity function, so that it doesn't really modify anything (which is what would be used in case of raw because no serialization/deserialization needed).

Also, I would like to avoid mutating attributes in different contexts like that if possible.

tchannel/thrift/serializer.py
```diff
@@ -0,0 +1,55 @@
...
+    for h in headers:
+        result[h[0]] = h[1]
+
+    return result
+
+
+def serialize_body(call_args):
+
+    trans = TTransport.TMemoryBuffer()
+    proto = TBinaryProtocol.TBinaryProtocolAccelerated(trans)
```
abhinav: I would just use `fastbinary` directly.

abhinav: fastbinary.encode_binary(call_args, (call_args.__class__, call_args.thrift_spec))

    fastbinary.decode_binary(result, TMemoryBuffer(body), (result_type, result_type.thrift_spec))

tchannel/tornado/broker.py
```diff
@@ -64,7 +64,11 @@ def handle_call(self, req, resp, proxy):
         return handler(req, resp, proxy)
 
     @tornado.gen.coroutine
-    def send(self, client, endpoint, header, body,
+    def send(self,
+             client,  # operation?
```
abhinav: what's this comment?

tchannel/schemes/json.py
breerly: @junchaowu yeah - since these are public API's I'm leaning towards re-stated the full method signatures.

tchannel/transport.py
```diff
@@ -0,0 +1,41 @@
...
+    return args
+
+
+class TransportHeaders(object):
+    """Transport Headers common between Request & Response"""
+
+    def __init__(self, scheme=None, failure_domain=None, **kwargs):
```
abhinav: what's with the kwargs

tchannel/schemes/json.py
breerly: Those are good points... might make sense to add request filters where the first one is a serialization filter.

tchannel/tchannel.py
```diff
@@ -0,0 +1,80 @@
...
+            arg3=arg3,
+            headers={'as': scheme}  # TODO this is nasty...
+            # TODO what about other transport headers?
+        )
+
+        # unwrap response
+        header = yield response.get_header()
+        body = yield response.get_body()
+        t = transport.to_kwargs(response.headers)
+        t = ResponseTransportHeaders(**t)
```
abhinav: This seems fragile. Don't expose the kwargs outside like that.

tchannel/schemes/json.py
breerly: I'll brew on it over the weekend.

tchannel/thrift/module.py
breerly: :+1:

tchannel/thrift/module.py
breerly: nice catch - i was slowly refactoring this stuff into something more manageable - that's totally what I'll do.

tchannel/schemes/__init__.py
abhinav: What? You're doing intentional circular imports? Surely there's a better way? Move the constants to another schemes/foo.py and import that everywhere?

tchannel/tornado/broker.py
breerly: me being confused by what this parameter is actually supposed to be lol

tests/thrift/test_server.py
```diff
@@ -24,7 +24,7 @@
 import pytest
 from thrift.Thrift import TType
 
-from tchannel.thrift.scheme import ThriftArgScheme
+from tchannel.dep.thrift_arg_scheme import DeprecatedThriftArgScheme
```
breerly: I'll get rid of this class when I get the server PR up

tchannel/tchannel.py
breerly: Let's talk before I merge about a better way :)

abhinav: Construct kwargs in this module? I mean, you're just trying to do a conversion from the `old_response.headers` dict to a `ResponseTransportHeaders` object. If you think that class should know that it's usually being built from old-style response headers, then you should have a `from_dict` or similar class method that just returns the object, not the constructor kwargs. If the class shouldn't know this detail, then the kwargs should be constructed manually in this module/class.

tchannel/schemes/__init__.py
breerly: hate to break the package w non-homogenous types :\

tchannel/transport.py
breerly: this allows nonsense kwargs to be passed in - i have it here because `to_kwargs` creates all possible transport headers, even though responsetransportheader is a subset

tchannel/__init__.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+from .tchannel import TChannel  # noqa
+from .thrift import from_thrift_module  # noqa
```
breerly: @abhinav @junchaowu @blampe last chance to suggest a better name for this

tchannel/event.py
```diff
@@ -45,6 +45,7 @@
     after_send_system_error=0x42,
     on_operational_error=0x50,
     on_operational_error_per_attempt=0x51,
+    on_application_error=0x52,
```
breerly: @abhinav this is weird... i didn't do this...

tchannel/schemes/__init__.py
breerly: will add TODO

tchannel/schemes/json.py
breerly: will add TODO

tchannel/thrift/module.py
breerly: will add TODO

tchannel/thrift/module.py
breerly: will add TODO

tchannel/thrift/module.py
breerly: will add TODO

tchannel/thrift/serializer.py
breerly: will add TODO

tchannel/__init__.py
abhinav: How about,

    thrift_request_builder(module=...)

abhinav: top level `from_*` doesn't really say what's being built.

tests/tornado/test_dispatch.py
```diff
@@ -79,3 +80,20 @@ def handler(req, response, proxy):
...
+
+    dispatcher.register('foo', handler)
+
+    yield dispatcher.handle_call(req, connection)
+
+    connection.tchannel.event_emitter.fire.assert_called_with(
+        EventType.on_application_error,
+        req,
+        mock.ANY,
+    )
```
breerly: didn't add this either :| - funked up the rebase :|

tchannel/schemes/thrift.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+        # raise application exception, if present
+        for exc_spec in request.result_type.thrift_spec[1:]:
+            exc = getattr(body, exc_spec[2])
+            if exc is not None:
+                raise exc
+
+        # success
+        response.body = body.success
```
abhinav: The `success` attribute won't exist on `body` if the result type was `void`. See what `client_for` does:

        if result_spec[0]:
            # Non-void return type. Return the result.
            success = getattr(call_result, result_spec[0][2])
            if success is not None:
                raise gen.Return(success)
        else:
            # No return type specified and no exceptions raised.
            raise gen.Return(None)
(`result_spec[0]` will be None when the return type was void.)

abhinav: Also, you want to raise some sort of exception if `result_spec[0]` was set (so a return value was expected), but `body.success` was None. Because that means that the server didn't return a value.

tchannel/tchannel.py
abhinav: ^

tchannel/thrift/module.py
abhinav: ^

tchannel/thrift/serializer.py
```diff
@@ -0,0 +1,55 @@
...
+
+def deserialize_headers(headers):
+
+    headers = io.BytesIO(headers)
+    headers = _headers_rw.read(headers)
+
+    result = {}
+
+    for h in headers:
+        result[h[0]] = h[1]
```
abhinav: result = dict(headers)

tchannel/transport.py
abhinav: Yeah, so ResponseTransportHeaders should `kwargs.pop` anything that the parent class doesn't expect.

tchannel/__init__.py
breerly: my thinking is that they would do something like:

```
donut_service = from_thrift_module(...)
```

That actually reads out well enough IMO

tchannel/response.py
breerly: will add TODO

tchannel/transport.py
breerly: Going to ship it this way - I don't see an immediate need to allow user-defined transport headers, especially since I want the response to end up in a well-defined object

tchannel/__init__.py
abhinav: I mean it doesn't say what's being built. "from_thrift_module" could be building a client, a request builder, a request handler, or anything else. The module name doesn't add any information on what's being build either.

    donut_service = thrift_request_builder(..)

Is equally readable, but it makes it clear that the new `donut_service` object builds requests for donut service. I would even say this makes sense:

    donut_service = ThriftRequestBuilder(...)

(Or Maker instead of Builder.)

My thought here is that there is more than one uses case for doing things dynamically given the module for a single Thrift service.

tchannel/transport.py
breerly: yeah... i buy that... will probably rework this whole area

junchaowu:
Considering our schedule, it looks good to me now. 
Implement those TODO later.

abhinav:
:shipit:

PR #73, opened 2015-08-03
add testing for streaming examples

blampe:
The API is broken right now but this asserts bytes can still be sent over the wire.

@abhinav @breerly @junchaowu


abhinav:
I thoughts streaming was broken?

blampe:
The API can pass bytes around but doesn't currently allow the use cases we're interested in (we don't get a chunk "streamed" to us until it fills up).

abhinav:
In that case I don't see what this test adds, but :ok:.

blampe:
We're going to be refactoring this stuff significantly and I want to make sure our examples stay relevant/working.

Issue #72, opened 2015-08-03
add an event hook for unhandled application errors


blampe:
When we log "An unexpected error has occurred from the handler" we should also fire off an event for unhandled exceptions.

PR #74, opened 2015-08-03
add an event hook for application errors

blampe:
refs #72 

@junchaowu @breerly @abhinav

### DIFF
--- a/tchannel/event.py
+++ b/tchannel/event.py
@@ -115,7 +115,19 @@ class EventHook(object):
         pass
 
     def on_application_error(self, request, err):
-        """Called on uncaught exceptions from request handlers."""
+        """Called on uncaught exceptions from request handlers.
+
+        :param request:
+            The :py:class:`tchannel.tornado.request.Request` object associated
+            with this uncaught exception.
+
+        :param err:
+            An instance of the unhandled exception.
+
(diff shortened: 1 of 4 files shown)


tests/test_event.py
```diff
@@ -44,11 +45,18 @@ def test_decorator_registration():
...
         called[0] = True
 
+    @registrar.before_send_request
+    def bar():
+        called[1] = True
+
     event_emitter.fire(EventType.before_send_request)
+
     assert called[0] is True
+    assert called[1] is True
```
abhinav: You can use mocks to assert something was called instead of this.

    hook1 = mock.Mock()
    registrar.before_send_request(hook1)
    hook1.assert_called_once_with()

tchannel/event.py
```diff
@@ -113,6 +114,10 @@ def on_operational_error_per_attempt(self, request, err):
         """Called after an operational error happens for each attempt."""
         pass
 
+    def on_application_error(self, request, err):
+        """Called on uncaught exceptions from request handlers."""
+        pass
+
```
abhinav: Document the parameters?

abhinav:
lgtm besides nits

tests/test_event.py
blampe: I'll keep it this way since it's intentionally testing the decoration.

PR #75, opened 2015-08-04
fix flakey example tests

blampe:
@breerly @abhinav @junchaowu


abhinav:
I restarted the tests and they just worked. Whoa.

:shipit:

PR #79, opened 2015-08-06
Improve tcurl doc

jcorbin:
- separate server run from ioloop start
- add listening print
- use an optional hostlist for advertise
- update tcurl invocation for new v4.10 options (-H and -t file)

r @abhinav @junchaowu


abhinav:
:+1:

PR #85, opened 2015-08-07
throw error when receiving no value from a non-void thrift procedure,…

### LINKED ISSUE #83: Fix missing Thrift error case
As @abhinav pointed out in `schemes/thrift.py`:

> if result_spec[0] was not None and no value was returned, that's a user-error and an exception needs to be raised

breerly:
… fixes #83

### DIFF
--- a/tchannel/schemes/thrift.py
+++ b/tchannel/schemes/thrift.py
@@ -5,6 +5,7 @@ from __future__ import (
 from tornado import gen
 
 from . import THRIFT
+from tchannel.errors import ValueExpectedError
 from tchannel.thrift import serializer
 
 
@@ -58,6 +59,15 @@ class ThriftArgScheme(object):
 
         # success - non-void
         if len(result_spec) >= 1 and result_spec[0] is not None:
+
+            # value expected, but got none
+            # TODO - server side should use this same logic
+            if body.success is None:
+                raise ValueExpectedError(
+                    'called a non-void procedure %s but received no value' %
+                    request.endpoint
+                )
+
             response.body = body.success
             raise gen.Return(response)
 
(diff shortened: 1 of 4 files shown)


tests/schemes/test_thrift.py
```diff
@@ -9,7 +9,7 @@
     TChannel, thrift_request_builder,
     schemes, response
 )
-from tchannel.errors import OneWayNotSupportedError
+from tchannel.errors import OneWayNotSupportedError, ValueExpectedError
```
blampe: Nit: one import per line.

tchannel/schemes/thrift.py
```diff
@@ -58,6 +59,15 @@ def __call__(self, request=None, headers=None, timeout=None,
 
         # success - non-void
         if len(result_spec) >= 1 and result_spec[0] is not None:
+
+            # value expected, but got none
+            # TODO - server side should use this same logic
+            if body.success is None:
+                raise ValueExpectedError(
+                    'called a non-void procedure %s but received no value' %
```
abhinav: How about changing this to: "Expected a value to be returned for %s but received None."

abhinav:
:+1:

tchannel/schemes/thrift.py
breerly: Yup

breerly: Actually... I like somehow indicating that non-void procedures need to return values... or that only void procedures can return None

blampe:
LGTM!

tests/schemes/test_thrift.py
breerly: @abhinav @junchaowu do you guys tend to do 1 import per-line?

PR #88, opened 2015-08-07
Request in tchannel.thrift should be required

breerly:
cc @blampe @abhinav

### DIFF
--- a/tchannel/schemes/thrift.py
+++ b/tchannel/schemes/thrift.py
@@ -18,8 +18,10 @@ class ThriftArgScheme(object):
         self._tchannel = tchannel
 
     @gen.coroutine
-    def __call__(self, request=None, headers=None, timeout=None,
+    def __call__(self, request, headers=None, timeout=None,
                  retry_on=None, retry_limit=None):
+
+        assert request, "a ThriftRequest is required"
 
         if headers is None:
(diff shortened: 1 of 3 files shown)


tchannel/schemes/thrift.py
```diff
@@ -18,9 +18,11 @@ def __init__(self, tchannel):
         self._tchannel = tchannel
 
     @gen.coroutine
-    def __call__(self, request=None, headers=None, timeout=None,
+    def __call__(self, request, headers=None, timeout=None,
                  retry_on=None, retry_limit=None):
 
+        assert request, "a ThriftRequest is required"
```
blampe: If you want to guard here this should be an isinstance check (I can pass `True`). I'd also recommend a different exception type.

breerly: Yeah... we shouldn't be throwing `AssertionError` on public API's. I'm going to fix all of them in one commit however.

breerly: Actually for this case an ABC makes sense since the isinstance will be cheap and we'll be able to know it's a thrift response.

abhinav: IMO don't do `isinstance` checks. In this case, just check for None (not truthiness).

PR #89, opened 2015-08-07
Raise ValueError when anything but map[string]string is used for Thrift

breerly:
cc @blampe @abhinav @junchaowu

### DIFF
--- a/tests/schemes/test_thrift.py
+++ b/tests/schemes/test_thrift.py
@@ -1144,41 +1144,24 @@ def test_value_expected_but_none_returned_should_error():
 
 
 @pytest.mark.gen_test
-@pytest.mark.callz
-def test_headers_should_be_a_map_of_strings():
+@pytest.mark.call
+@pytest.mark.parametrize('headers', [
+    {'key': 1},
+    {1: 'value'},
+    {'key': {'key': 'value'}},
+    100,
+    -100,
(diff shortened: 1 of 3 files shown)


tests/schemes/test_thrift.py
```diff
@@ -1141,3 +1141,44 @@ def testString(request, response, proxy):
...
+@pytest.mark.gen_test
+@pytest.mark.callz
+def test_headers_should_be_a_map_of_strings():
+
+    tchannel = TChannel('client')
+
+    with pytest.raises(ValueError):
+        yield tchannel.thrift(
+            request=True,
+            headers={'key': 1},
```
blampe:
```
@pytest.mark.parametrize("headers", [
    {'key': 1},
    ...
])
def test_headers_should_be_a_map_of_strings(headers):
    with ...:
        yield tchannel.thrift(
            ...
            headers=headers
        )
```

junchaowu: +1

abhinav: +1

abhinav:
lgtm besides test parameterization

tests/schemes/test_thrift.py
breerly: yup

PR #97, opened 2015-08-11
remove broker.py

junchaowu:
Remove broker.py, scheme.py, deprecated thrift arg scheme.py. Some places is written for backward compatibility purpose.
Fix style issue in the code.
Lots of tests fixing. No breaking api change.
No stackcontext in this diff.


examples/benchmark/thrift/service/KeyValue.py
```diff
@@ -26,10 +26,15 @@
...
+from thrift.protocol.TBase import TBase
+from thrift.protocol.TBase import TExceptionBase
+from thrift.Thrift import TApplicationException
+from thrift.Thrift import TException
+from thrift.Thrift import TMessageType
 from thrift.Thrift import TProcessor
-from thrift.protocol.TBase import TBase, TExceptionBase
+from thrift.Thrift import TType
+
+from ttypes import *
```
breerly: @junchaowu this is generated code dood

examples/benchmark/thrift/service/constants.py
```diff
@@ -26,6 +26,9 @@
...
 #
 
-from thrift.Thrift import TType, TMessageType, TException, TApplicationException
-from ttypes import *
+from thrift.Thrift import TApplicationException
+from thrift.Thrift import TException
+from thrift.Thrift import TMessageType
+from thrift.Thrift import TType
 
+from ttypes import *
```
breerly: this is generated code d00d

tchannel/retry.py
```diff
@@ -1,6 +1,7 @@
-from __future__ import (
-    absolute_import, division, print_function, unicode_literals
-)
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
```
breerly: hmmm.... not sure how I feel about dis

tchannel/serializer/json.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import absolute_import
+
+import json
+
+from tchannel.schemes import JSON
+
+
+class JsonSerializer(object):
+    def __init__(self):
+        self.name = JSON
```
breerly: Why are you duplicating this? This constant already exists in `tchannel.schemes`...

tchannel/serializer/raw.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import
+
+from tchannel.schemes import RAW
+
+
+class RawSerializer(object):
+    def __init__(self):
```
breerly: These don't have any state... please remove this class in favor of regular old functions

breerly:
@junchaowu can you back out all the auto-formatting? It makes this PR a pain to read

tests/integration/json/test_json_server.py
```diff
@@ -92,22 +90,21 @@ def json_server():
...
+
     header = sample_json
     body = sample_json
-    resp = yield ArgSchemeBroker(JsonArgScheme()).send(
-        client,
-        endpoint,
-        header,
-        body,
+    resp = yield tchannel.json(
+        service='endpoint1',
```
breerly: endpoint1???

tchannel/tornado/tchannel.py
```diff
@@ -61,12 +63,6 @@ class TChannel(object):
 
     FALLBACK = RequestDispatcher.FALLBACK
 
-    _SCHEMES = {
-        'raw': scheme.RawArgScheme,
-        'json': scheme.JsonArgScheme,
-        # 'http': scheme.HttpArgScheme, TODO
-    }
```
breerly: woohoo!

tests/schemes/test_thrift.py
```diff
@@ -341,7 +344,7 @@ def testStruct(request, response, proxy):
         # TODO server getting headers in non-friendly format,
         # create a top-level request that has friendly headers :)
         # assert request.headers == {'req': 'headers'}
-        assert request.headers == [['req', 'header']]
+        assert request.headers == {'req': 'header'}
```
breerly: does this not break? I'm looking to fix this in my PR

tests/schemes/test_thrift.py
```diff
@@ -1162,6 +1163,6 @@ def test_headers_should_be_a_map_of_strings(headers):
 
     with pytest.raises(ValueError):
         yield tchannel.thrift(
-            request=True,
+            request=mock.MagicMock(),
```
breerly: is that really needed?

tests/schemes/test_thrift.py
junchaowu: Already rebased. fixed the merge conflict.

tests/schemes/test_thrift.py
junchaowu: This original one breaks test since the thrift method will access request.result_type.  If you pass a True, it will break.

tchannel/serializer/raw.py
junchaowu: The thrift serializer needs arg_type and result_type class in order to do deserializer. json, raw don't need that but in order to keep three look same.  I make three all this way

tchannel/serializer/json.py
junchaowu: I need to compare the endpoint scheme name with the value in the 'as' key - word in the transport header.

breerly: Please move to a class constant then.

tchannel/serializer/raw.py
breerly: Hmmm... that's pretty smelly - I'll bet we can rework it to avoid state - serialization should be pure functions

tests/schemes/test_thrift.py
breerly: I think we expect ValueError to be raised before request would be accessed

tchannel/tornado/dispatch.py
```diff
@@ -21,20 +21,22 @@
...
 from tornado import gen
 
 from ..errors import InvalidEndpointError
-from ..errors import InvalidMessageError
 from ..errors import TChannelError
 from ..event import EventType
 from ..handler import BaseRequestHandler
 from ..messages.error import ErrorCode
-from .broker import ArgSchemeBroker
+from ..serializer.raw import RawSerializer
```
breerly: this should be 

```
from tchannel.serializer.raw import ...
```

Upwards relative imports hamper refactorability.

tchannel/serializer/json.py
junchaowu: sure

tchannel/tornado/dispatch.py
junchaowu: We had this discussion before. Look at other imports. If just import from one level before, we still use relative import

tests/schemes/test_thrift.py
breerly: What did you change so that headers started coming in this format? You'll want to record it as a breaking change in the changelog and detail the upgrade path in upgrade.rst.

tchannel/tornado/dispatch.py
breerly: Fair enough - I'll submit a PR making a case for it across the board.

tchannel/serializer/thrift.py
```diff
@@ -0,0 +1,62 @@
...
+
+    _headers_rw = rw.headers(
+        rw.number(2),
+        rw.len_prefixed_string(rw.number(2)),
+        rw.len_prefixed_string(rw.number(2)),
+    )
+
+    def __init__(self, deserialize_type):
+        self.name = THRIFT
+        self.deserialize_type = deserialize_type
```
breerly: This is only used in deserialize_body, make it a parameter to that method, and then you can make all these serializers pure functions. In fact, the serialize_header, then serialize_body thing is a bit unessecary, since you'll always want to perform these operations at the same time. Suggest:

```py
from serializers.json import serialize, deserialize

headers, body = serialize(headers, body)

#...

headers, body = deserialize(headers, body)
```

examples/simple/thrift/client.py
```diff
@@ -1,9 +1,11 @@
 import json
```
abhinav: Oh it looks like some of the files don't have license blurbs anymore.

Install the hook by running `sh scripts/install-hooks.sh` in the project root please. @breerly, you too.

tchannel/schemes/json.py
```diff
@@ -1,12 +1,12 @@
-from __future__ import (
-    absolute_import, division, print_function, unicode_literals
-)
-
-import json
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
```
abhinav: You need to reconfigure your auto-formatter

tchannel/serializer/raw.py
breerly: @junchaowu I put a note elsewhere on how to fix - search "This is only used in deserialize_body"

examples/simple/thrift/client.py
breerly: @abhinav are we sure that it's a requirement to have the license at the top of each file? I feel like with that and long doc blocks, it's really hard to get to the actual implementation :)

tchannel/schemes/json.py
breerly: Yeah this is gonna be a trend otherwise

tchannel/serializer/json.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def deserialize_header(self, obj):
+        if not obj:
+            return None
+        return json.loads(obj)
+
+    def serialize_header(self, obj):
+        if not obj:
+            obj = {}
```
abhinav: Did we decide that `{}` was the default value for headers and body? I don't get why `tchannel.json(body)` would also send `"{}"` even though I haven't set anything there.

tchannel/serializer/json.py
```diff
@@ -0,0 +1,30 @@
...
+        return json.dumps(obj)
+
+    def deserialize_body(self, obj):
+        if not obj:
+            return None
+        return json.loads(obj)
+
+    def serialize_body(self, obj):
+        if not obj:
+            obj = {}
```
abhinav: Same as above. 

Also, we allow arbitrary JSON-like objects. I could pass an empty string there. That's a valid object that can be converted to JSON, but this will convert my `""` to `"{}"`.

tchannel/serializer/json.py
```diff
@@ -0,0 +1,30 @@
...
+        return json.dumps(obj)
+
+    def deserialize_body(self, obj):
+        if not obj:
+            return None
+        return json.loads(obj)
+
+    def serialize_body(self, obj):
+        if not obj:
```
blampe: What's up with the defaults here? 

`serialize_body(None) -> {}` (instead of "null"?)

`serialize_body(False) -> {}` (instead of "false"?)

`serialize_body(True) -> "true"` (instead of "{}"?)

tchannel/serializer/thrift.py
```diff
@@ -0,0 +1,62 @@
...
+    def deserialize_header(self, headers):
+        headers = io.BytesIO(headers)
+        headers = self._headers_rw.read(headers)
+        result = dict(headers)
+
+        return result
+
+    def serialize_body(self, call_args):
+
+        # TODO - use fastbinary directly
```
blampe: Nah we'll just use @abhinav's magic.

tchannel/serializer/raw.py
abhinav: +1 on that.

`name` shouldn't be an instance attribute. It doesn't change in any of the cases.

tchannel/tornado/dispatch.py
```diff
@@ -82,21 +85,28 @@ def handle_call(self, request, connection):
             request,
         )
 
-        endpoint = self.endpoints[request.endpoint]
-
+        handler = self.handlers[request.endpoint]
+        if not request.headers.get('as', None) == handler.req_serializer.name:
```
abhinav: if not a == b

Change to

    if a != b

tchannel/tornado/dispatch.py
```diff
@@ -82,21 +85,28 @@ def handle_call(self, request, connection):
...
+            raise gen.Return(connection.send_error(
+                ErrorCode.bad_request,
+                "Invalid arg scheme in request header",
+                request.id,
```
abhinav: `send_error` returns a future. You have to yield it. You can't return it.

abhinav: And then you need a separate `return` statement.

(You can do `return` without any arguments inside a generator.)

tchannel/tornado/dispatch.py
```diff
@@ -160,27 +165,24 @@ def check_health(request, response, proxy):
             A function that gets called with ``Request``, ``Response``, and
             the ``proxy``.
 
-        :param broker:
-            Broker injects customized serializer and deserializer into
-            request/response object.
+        :param req_serializer:
+            Arg scheme serializer of this endpoint. It should be
+            ``RawSerializer``, ``JsonSerializer``, and ``ThriftSerializer``.
```
abhinav: Well, I mean it doesn't have to be one of those three. As long as it satisfies the `Serializer` interface, it's a valid argument here.

Actually, can you add a `Serializer` class that just defines the interface (`(de)serialize_{headers,body}`)? You can make the Raw, JSON, and Thrift serializers inherit it.

tchannel/serializer/json.py
breerly: I don't feel strongly either way - so long as the client-side and server-side experience is ok. Also, if I remember correctly, removing `{}` will break tests, but I didn't look into why that was.

tchannel/serializer/json.py
breerly: Passing empty strings is not an acceptable experience - they will either pass a dict-like thing or nothing at all.

tchannel/serializer/thrift.py
breerly: I explored using `fastbinary`, but it doesn't get us away from relying on the apache generated code... so I agree that we should just wait until `thriftrw` is ready :)

tchannel/serializer/json.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import absolute_import
+
+import json
+
+from tchannel.schemes import JSON
+
+
+class JsonSerializer(object):
+    def __init__(self):
```
abhinav: BTW consider adding [`__slots__`](https://docs.python.org/2/reference/datamodel.html#slots) to the `Serializer` classes. Since we'll be instantiating these twice for each request/response (assuming we don't just make them stateless and re-use the same ones), we should avoid allocating dictionaries for each instance. `__slots__` will take care of that.

tchannel/tornado/request.py
```diff
@@ -92,7 +94,7 @@ def rewind(self, id=None):
         self.tracing = Trace()
 
     @property
-    def arg_scheme(self):
+    def arg_serializer(self):
         return self.headers.get('as', None)
```
abhinav: What? This gets the arg scheme name, not the serializer.

tchannel/tornado/request.py
```diff
@@ -112,10 +114,10 @@ def get_header(self):
         :return: a future contains the deserialized value of header
         """
         raw_header = yield get_arg(self, 1)
-        if not self.scheme:
+        if not self.serializer:
```
abhinav: You gave it a default of `RawSerializer`. There won't be any case where `serializer` will be None -- unless the user makes it break manually, in which case this should blow up anyway.

tchannel/tornado/response.py
```diff
@@ -124,10 +125,10 @@ def get_body(self):
         """
 
         raw_body = yield get_arg(self, 2)
-        if not self.scheme:
+        if not self.serializer:
             raise tornado.gen.Return(raw_body)
```
abhinav: Same as above

tchannel/serializer/json.py
breerly: @abhinav there's no need to use classes here at all - they should be pure functions

abhinav:
I'm seeing a couple new files but no new tests. Please add tests that verify the behavior of `*Serializer` classes.

tchannel/serializer/thrift.py
abhinav: I asked for fastbinary to be used because this is how the generated code looks for `read/write`:

    def write(self, oprot):
      if oprot.__class__ == TBinaryProtocol.TBinaryProtocolAccelerated and self.thrift_spec is not None and fastbinary is not None:
        oprot.trans.write(fastbinary.encode_binary(self, (self.__class__, self.thrift_spec)))
        return
      [...]

    def read(self, iprot):
      if iprot.__class__ == TBinaryProtocol.TBinaryProtocolAccelerated and isinstance(iprot.trans, TTransport.CReadableTransport) and self.thrift_spec is not None and fastbinary is not None:
        fastbinary.decode_binary(self, iprot.trans, (self.__class__, self.thrift_spec))
        return
      [...]

Instantiating the buffer, the protocol, etc. is unnecessary work when we can just do that.

tchannel/serializer/json.py
abhinav: No, I'm saying that the default of `{}` for false-like values is wrong. `call(""/False/[])` should not send `"{}"`

tchannel/tornado/request.py
breerly: this should be `scheme` then

tchannel/serializer/json.py
abhinav: I mean, there's no reason they can't be grouped together in a `Serializer` class. Instances of that get passed around, and consumers don't care how they were instantiated or whether they carry state. Like, if I wanted to implement a  generic serializer that wraps another serializer in an envelope of some kind, the top-level function approach would leave me unable to do that without serious refactoring.

tchannel/serializer/json.py
junchaowu: It basically matches what is in our new client design. Doesn't matter which way we choose as long as server and client are consistent.

tchannel/serializer/raw.py
breerly: Ok - because you need to treat all these types the same and pass them around - let's keep them as classes for now @junchaowu @abhinav

tchannel/serializer/json.py
abhinav: Except they're not consistent. My client sent no headers, or an empty list as headers, and my server received an empty dict.

tests/schemes/test_thrift.py
junchaowu: We still expect ValueError. Just because I change the order of the code so we use mock here.

tchannel/serializer/thrift.py
breerly: sure why not :)

tchannel/thrift/client.py
```diff
@@ -163,18 +164,19 @@ def send(self, *args, **kwargs):
...
+        ).send(
             endpoint,
-            {},
-            call_args,  # body
-            protocol_headers=self.protocol_headers,
+            header,
+            body,  # body
+            headers=self.protocol_headers,
             traceflag=self.trace
         )
```
breerly: can you specify all the kwargs here?

tchannel/tornado/dispatch.py
breerly: What do you guys think of just:

```
headers, body = serializer.serialize()
```

I can't think of a use-case that you wouldn't want to call these at the same time.

cc @abhinav @junchaowu

tests/schemes/test_thrift.py
breerly: k

tchannel/tornado/dispatch.py
breerly: Actually - I disagree with this. These need to be separate operations so they can fail separately.

breerly:
:shipit: just fix tests and address nits where you see fit

examples/simple/thrift/client.py
breerly: update - turns out this is required

PR #102, opened 2015-08-11
Simplify event types (1)

junchaowu:
This is first patch of a serial patches about refactoring events and retry.
The first one is intended to remove unnecessary event.

### DIFF
--- a/tchannel/event.py
+++ b/tchannel/event.py
@@ -40,12 +40,9 @@ EventType = enum(
     after_receive_request=0x21,
     before_receive_response=0x30,
     after_receive_response=0x31,
-    after_receive_system_error=0x40,
-    after_receive_system_error_per_attempt=0x41,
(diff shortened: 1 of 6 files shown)


tchannel/event.py
```diff
@@ -40,12 +40,9 @@
...
     after_receive_response=0x31,
-    after_receive_system_error=0x40,
-    after_receive_system_error_per_attempt=0x41,
-    after_send_system_error=0x42,
-    on_operational_error=0x50,
-    on_operational_error_per_attempt=0x51,
-    on_application_error=0x52,
+    after_receive_error=0x41,
+    after_send_error=0x42,
+    on_exception=0x50,
```
abhinav: Changelog.  You need to add what you removed/renamed/combined.

abhinav:
Update the changelog and :+1:

PR #104, opened 2015-08-12
Remove unused hierarchy

breerly:
These don't need to exist, so they shouldn't.


tchannel/handler.py
```diff
@@ -19,118 +19,8 @@
...
-            Connection through which the call was made
-        :returns:
-            Nothing. The response must be sent using the
-            implementation-specific connection object.
-        """
-        raise NotImplementedError("Must be implemented.")
-
-
-class CallableRequestHandler(RequestHandler):
+class CallableRequestHandler(object):
```
abhinav: Update docstring

tchannel/tornado/dispatch.py
```diff
@@ -30,15 +30,19 @@
...
+from ..errors import StreamingError
+from ..messages import Types
+
+
 Handler = namedtuple('Handler', 'endpoint req_serializer resp_serializer')
 
 
-class RequestDispatcher(BaseRequestHandler):
+class RequestDispatcher(object):
     """A synchronous RequestHandler that dispatches calls to different
```
abhinav: Update docstring

tchannel/tornado/dispatch.py
```diff
@@ -57,11 +61,63 @@ def my_method(request, response, proxy):
     FALLBACK = object()
 
     def __init__(self):
```
abhinav: `__slots__`

abhinav:
:+1:

PR #109, opened 2015-08-12
allow ctrl-c'ing benchmark examples

blampe:
This makes them more compatible with plop.

@abhinav @junchaowu @breerly


abhinav:
:+1:

PR #112, opened 2015-08-13
retry refactor (3)

junchaowu:
Several major changes on this diff.
1. move the peer select logic to the 'send' method. Whenever tchannel calls send, it will pick a new peer based on current peers and connections. Before we have a fix peer assigned to the peerclientoperation class. We may be in trouble that the peer is no long healthy but the peerclientoperation instance still uses it. It also makes more sense on the retry logic.
2. Refactor events emit code. Previously we will have a potential bug that we record an send_request event, but due to retry happening, the response that we finally received is not matched to the first request. Because during retry, after we switch to a new peer and a new connection, the request's id, connection host, port (used in the tracing) will become different. (sorry terrible explanation, it will be better explain with code) 
3. Optimized retry code a bit. Retry workarounds.


junchaowu:
add changelog after all changes got approved

tchannel/tornado/peer.py
```diff
@@ -540,121 +551,100 @@ def send(
...
     def _send(self, connection, req):
+        # event: send_request
+        self.tchannel.event_emitter.fire(EventType.before_send_request, req)
         response_future = connection.send_request(req)
+
         with timeout(response_future, req.ttl):
-            response = yield response_future
+            try:
+                response = yield response_future
+            except ProtocolError as protocol_error:
```
blampe: Naming issue here: ProtocolError is reserved for a very specific kind of error that actually tears down our connection to the peer.

blampe:
> Before we have a fix peer assigned to the peerclientoperation class. We may be in trouble that the peer is no long healthy but the peerclientoperation instance still uses it. It also makes more sense on the retry logic.

If we're fixing a bug we should write a test for it.

blampe:
> Previously we will have a potential bug that we record an send_request event

Same goes for that.

tchannel/tchannel.py
```diff
@@ -132,7 +132,7 @@ def call(self, scheme, service, arg1, arg2=None, arg3=None,
             arg2=arg2,
             arg3=arg3,
             headers=transport_headers,
-            attempt_times=retry_limit,
+            retry_limit=retry_limit,
```
abhinav: So I remember that we decided to call this attempt_times because this doesn't specify the number of times you'll retry but the number of attempts you'll make.

That is, I would think that `retry_count=0` means I'll make one attempt but no retries. In this case, `retry_count=0` means that you'll make no attempts or retries.

I'm okay with dropping `attempt_times` but then you have to change the behavior of `retry_limit` to only limit the number of retries, and always perform at least one attempt.

tchannel/tornado/peer.py
junchaowu: Currently ProtocolError represent Error frames we received from network. We use code to know the specific error message. We will make different Errors though

tchannel/tornado/hyperbahn.py
```diff
@@ -64,7 +64,7 @@ def _advertise(tchannel, service):
                 ]
             }),
             headers={'as': 'json'},
-            attempt_times=1,
+            retry_limit=1,
```
abhinav: Is this supposed to be 1 under the new definition of `retry_limit`?

tchannel/tornado/peer.py
```diff
@@ -491,24 +495,35 @@ def send(
...
+        if not peer:
+            raise NoAvailablePeerError("Can't find available peer.")
+        connection = yield peer.connect()
+
         arg1, arg2, arg3 = (
             maybe_stream(arg1), maybe_stream(arg2), maybe_stream(arg3)
         )
 
-        attempt_times = attempt_times or DEFAULT_RETRY_LIMIT
+        retry_limit = retry_limit or DEFAULT_RETRY_LIMIT
```
abhinav: DEFAULT_RETRY_LIMIT will have to be changed?

tchannel/tornado/peer.py
```diff
@@ -540,121 +552,99 @@ def send(
...
-                                  self.service),
+                endpoint=Endpoint(peer.host, peer.port, self.service),
                 traceflags=traceflag,
             )
         )
 
         # only retry on non-stream request
         if request.is_streaming_request or self._hostport:
-            attempt_times = 1
+            retry_limit = 1
```
abhinav: This needs to change under the new meaning?

tchannel/tornado/peer.py
```diff
@@ -540,121 +552,99 @@ def send(
...
 
         self.clean_up_outgoing_request(request, connection, protocol_error)
         if not self.should_retry(request, protocol_error,
-                                 num_of_attempt, max_attempt_times):
+                                 num_of_attempt, max_retry_limit):
             raise gen.Return((None, None))
 
-        result = yield self.prepare_next_request(request, peer, blacklist)
+        result = yield self.prepare_next_request(request, blacklist)
         raise gen.Return(result)
```
abhinav: this method shouldn't be a coroutine

tests/integration/test_retry.py
```diff
@@ -88,7 +97,7 @@ def test_retry_timeout():
                     're': retry.CONNECTION_ERROR_AND_TIMEOUT
                 },
                 ttl=0.005,
-                attempt_times=3,
+                retry_limit=3,
```
abhinav: All these values need to be changed if you're changing the meaning of the argument.

abhinav:
If you're changing the behavior and definition of `retry_limit`/`attempt_times`, you need to update things that relied on the old behavior.

tchannel/tchannel.py
```diff
@@ -132,7 +132,7 @@ def call(self, scheme, service, arg1, arg2=None, arg3=None,
             arg2=arg2,
```
abhinav: BTW have a look at the retry_limit docstring on line 91. It says it's a string. Should be int?

Also, finally, that docstring is actually correct. Before this, it was wrong (since we were doing num_attempts with retry_limit).

junchaowu: lol

tchannel/tornado/peer.py
```diff
@@ -540,121 +552,99 @@ def send(
...
     def _send(self, connection, req):
+        # event: send_request
+        self.tchannel.event_emitter.fire(EventType.before_send_request, req)
         response_future = connection.send_request(req)
-        with timeout(response_future, req.ttl):
-            response = yield response_future
 
+        with timeout(response_future, req.ttl):
+            try:
+                response = yield response_future
```
blampe: FWIW this method is a good example of an unnecessary coroutine - these are just some `add_done_callbacks`.

blampe:
:+1:

PR #114, opened 2015-08-14
improvments for exception handling

blampe:
Please pay attention to the hierarchy in `errors.py` and the exceptions we're now catching in tests. Compare these against [the spec](http://tchannel.readthedocs.org/en/latest/protocol/#code1_1).

* Get rid of `ProtocolError`. That's a very particular error (`0xff`) that we don't want everything subclassing. 
* Add `AlwaysRetryableError`, `PossiblyRetryableError`, and `UnretryableError` to make the individual exceptions more actionable.
* Got rid of some of the other exception classes like `TChannelApplicationError` (unused), `ConnectionClosedError` (already captured by `NetworkError`), `InvalidEndpointError ` (already captured by `BadRequest`), etc. 
* We were catching `InvalidChecksumError` and `StreamingError` and always returning those as `BadRequest`, now we return the original error.
* Switched the default handler to not use defaultdict because it's slow.
* I'm duplicating error codes (they're already in the message definition) but I didn't want this file importing from messages because it shouldn't have any dependencies.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,10 @@ Changelog
   the new server API. See the upgrade guide for details.
 - **BREAKING** - remove ``retry_delay`` option in the ``tchannel.tornado.send``
   method.
+- **BREAKING** - error types have been reworked significantly. In particular,
+  the all-encompassing ``ProtocolError`` has been replaced with more
+  granualar/actionable exceptions. See the upgrade guide for more info.
+
 
 
 0.15.2 (2015-08-07)
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -39,6 +39,27 @@ From 0.15 to 0.16
 
   After: no more ``retry_delay`` in  ``tchannel.tornado.TChannel.request.send()``
 
+- If you were catching ``ProtocolError`` you will need to catch a more specific
+  type, such as ``TimeoutError``, ``BadRequestError``, ``NetworkError``,
+  ``UnhealthyError``, or ``UnexpectedError``.
+
+- If you were catching ``AdvertiseError``, it has been replaced by
+  ``TimeoutError``.
+
+- If you were catching ``BadRequest``, it may have been masking checksum errors
+  and fatal streaming errors. These are now raised as ``FatalProtocolError``,
+  but in practive should not need to be handled when interacting with a
+  well-behaved TChannel implementation.
+
+- ``TChannelApplicationError`` was unused and removed.
+
+- Three error types have been introduced to simplify retry handling:
+  ``UnretryableError`` (for requests should never be retried),
+  ``AlwaysRetryableError`` (for requests that are always safe to retry), and
+  ``PossiblyRetryableError`` (for requests that are safe to retry on idempotent
+  endpoints).
+
+
 From 0.14 to 0.15
 -----------------
 
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -55,9 +55,9 @@ From 0.15 to 0.16
 - ``TChannelApplicationError`` was unused and removed.
 
 - Three error types have been introduced to simplify retry handling:
-  ``UnretryableError`` (for requests should never be retried),
-  ``AlwaysRetryableError`` (for requests that are always safe to retry), and
-  ``PossiblyRetryableError`` (for requests that are safe to retry on idempotent
+  ``NotRetryableError`` (for requests should never be retried),
+  ``RetryableError`` (for requests that are always safe to retry), and
+  ``MaybeRetryableError`` (for requests that are safe to retry on idempotent
   endpoints).
 
 
--- a/tchannel/errors.py
+++ b/tchannel/errors.py
@@ -70,51 +70,51 @@ class TChannelError(Exception):
         }[code](**kw)
 
 
-class AlwaysRetryableError(TChannelError):
+class RetryableError(TChannelError):
     pass
 
 
-class PossiblyRetryableError(TChannelError):
+class MaybeRetryableError(TChannelError):
     pass
 
 
-class UnretryableError(TChannelError):
+class NotRetryableError(TChannelError):
     pass
 
 
-class TimeoutError(PossiblyRetryableError):
+class TimeoutError(MaybeRetryableError):
     code = TIMEOUT
 
 
-class CanceledError(UnretryableError):
+class CanceledError(NotRetryableError):
     code = CANCELED
 
 
-class BusyError(AlwaysRetryableError):
+class BusyError(RetryableError):
     code = BUSY
 
 
-class DeclinedError(AlwaysRetryableError):
+class DeclinedError(RetryableError):
     code = DECLINED
 
 
-class UnexpectedError(PossiblyRetryableError):
+class UnexpectedError(MaybeRetryableError):
     code = UNEXPECTED_ERROR
 
 
-class BadRequestError(UnretryableError):
+class BadRequestError(NotRetryableError):
     code = BAD_REQUEST
 
 
-class NetworkError(PossiblyRetryableError):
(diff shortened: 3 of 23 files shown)


blampe:
@junchaowu @breerly @abhinav

abhinav:
:+1: 

The UPGRADE.rst updates for this will be fun.

breerly:
This affects customers, be sure to update 'ze changelog.

tchannel/errors.py
```diff
@@ -21,13 +21,20 @@
...
 
+TIMEOUT = 0x01
+CANCELED = 0x02
+BUSY = 0x03
+DECLINED = 0x04
+UNEXPECTED_ERROR = 0x05
+BAD_REQUEST = 0x06
+NETWORK_ERROR = 0x07
+UNHEALTHY = 0x08
+FATAL = 0xFF
```
breerly: Nice :)

tchannel/errors.py
```diff
@@ -35,97 +42,107 @@ class ProtocolError(TChannelError):
...
+            TIMEOUT: TimeoutError,
+            CANCELED: CanceledError,
+            BUSY: BusyError,
+            DECLINED: DeclinedError,
+            UNEXPECTED_ERROR: UnexpectedError,
+            BAD_REQUEST: BadRequestError,
+            NETWORK_ERROR: NetworkError,
+            UNHEALTHY: UnhealthyError,
+            FATAL: FatalProtocolError,
+        }[code](**kw)
```
breerly: plz move this dict to `ALL` under the glossary above in case it needs to be reused

tchannel/errors.py
```diff
@@ -35,97 +42,107 @@ class ProtocolError(TChannelError):
...
+
+
+class AlwaysRetryableError(TChannelError):
+    pass
+
 
-class InvalidMessageError(TChannelError):
-    """Represent an invalid message."""
+class PossiblyRetryableError(TChannelError):
```
breerly: Not sure how I feel about this... what's the use-case?

breerly:
Please move the tests related to `tchannel/errors.py` to `tests/test_errors.py`.

tchannel/errors.py
```diff
@@ -35,97 +42,107 @@ class ProtocolError(TChannelError):
...
     pass
 
 
-class InvalidEndpointError(TChannelError):
-    """Represent an message containing invalid endpoint."""
+class UnretryableError(TChannelError):
```
breerly: Every exception that is woken up from a protocol error should inherit from `ProtocolError`

tchannel/errors.py
blampe: Idempotency, like maybe I don't want to retry that ``sendEmail`` request. "Everything should be idempotent" but in practice that's easier said than done :)

tchannel/errors.py
blampe: It shouldn't, protocol error is a [very specific error case](http://tchannel.readthedocs.org/en/latest/protocol/#type1).

tchannel/errors.py
breerly: The naming on this is really awkward... @abhinav can you think of something more obvious? I can already imagine seeing client code where people are catching this and thinking wtf

tchannel/errors.py
breerly: Then there is a gap in my understanding - can we jam in person before you land this?

tchannel/errors.py
abhinav: I agree that 'always retryable' and 'possibly retryable' could be confusing; especially because always retryable sounds like immediately and infinitely retryable. 

How about just Retryable and with some sort of property that is true if the original request might have made it to the server?

breerly: Suggest `MaybeRetryableError`

tchannel/errors.py
```diff
@@ -35,97 +42,107 @@ class ProtocolError(TChannelError):
...
+            DECLINED: DeclinedError,
+            UNEXPECTED_ERROR: UnexpectedError,
+            BAD_REQUEST: BadRequestError,
+            NETWORK_ERROR: NetworkError,
+            UNHEALTHY: UnhealthyError,
+            FATAL: FatalProtocolError,
+        }[code](**kw)
+
+
+class AlwaysRetryableError(TChannelError):
```
breerly: Suggest `RetryableError`

tchannel/errors.py
breerly: Suggest `NotRetryableError`.

UPGRADE.rst
```diff
@@ -39,6 +39,27 @@ From 0.15 to 0.16
...
+  but in practive should not need to be handled when interacting with a
+  well-behaved TChannel implementation.
+
+- ``TChannelApplicationError`` was unused and removed.
+
+- Three error types have been introduced to simplify retry handling:
+  ``UnretryableError`` (for requests should never be retried),
+  ``AlwaysRetryableError`` (for requests that are always safe to retry), and
+  ``PossiblyRetryableError`` (for requests that are safe to retry on idempotent
+  endpoints).
```
junchaowu: The purpose to add these is to classify errors?

blampe: Similar to 4XX vs 5XX categories. I might not care that this was a failure to connect or a declined request, if it's safe to retry then I probably want to retry.

tchannel/tornado/peer.py
```diff
@@ -618,7 +617,8 @@ def send_with_retry(self, request, peer, retry_limit, connection):
             try:
                 response = yield self._send(connection, request)
                 raise gen.Return(response)
-            except (ProtocolError, TimeoutError) as error:
+            # Why are we retying on all errors????
+            except TChannelError as error:
```
junchaowu: There will be a problem here. Existing code use the ``code`` field in the ProtocolError to know if it is retry or not. If the new error type doesn't have ``code`` field, it will break.

blampe: The exception classes still have a `code` attribute on them if you need it.

The question here was because we're catching `ProtocolError` which all of our exceptions were inheriting from... it's still preserving that behavior by catching `TChannelError` (which is what everything now inherits from).

junchaowu: Yes. The logic in those place will also need to be updated.

blampe: What do you mean? This is already a breaking change, what is the update you're talking about? What should `UPGRADE.rst` tell consumers to do?

junchaowu: I mean internally. for example since we not have different error type, ideally doesn't need the code to tell which one to retry.

blampe: I see what you're saying, yeah we don't need to do this `should_retry_on_error` stuff anymore.

CHANGES.rst
```diff
@@ -19,7 +19,11 @@ Changelog
   the new server API. See the upgrade guide for details.
 - **BREAKING** - remove ``retry_delay`` option in the ``tchannel.tornado.send``
   method.
-- **BREAKING** - remove third arugment ``proxy`` from the server handler interface.
+- **BREAKING** - error types have been reworked significantly. In particular,
+  the all-encompassing ``ProtocolError`` has been replaced with more
+  granualar/actionable exceptions. See the upgrade guide for more info.
+- **BREAKING** - Remove third ``proxy`` argument from the server handler
```
breerly: Weird rebase?

blampe: Fixed the wording.

blampe:
How does everyone feeling about this?

tchannel/errors.py
abhinav: +1. Move dict to top-level constant.

abhinav:
minor nit. looks good otherwise

PR #122, opened 2015-08-18
Context

junchaowu:
Add stack context into tchannel. 
Removed the proxy class.
Make zipkintracehook not default.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ Changelog
   the new server API. See the upgrade guide for details.
 - **BREAKING** - remove ``retry_delay`` option in the ``tchannel.tornado.send``
   method.
+- **BREAKING** - remove third arugment ``proxy`` from the server handler interface.
+- **BREAKING** - make the zipkintracehook not default registered.
 
 
 0.15.2 (2015-08-07)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,7 +20,7 @@ Changelog
 - **BREAKING** - remove ``retry_delay`` option in the ``tchannel.tornado.send``
   method.
 - **BREAKING** - remove third arugment ``proxy`` from the server handler interface.
-- **BREAKING** - make the zipkintracehook not default registered.
+- **BREAKING** - ``ZipkinTraceHook`` is not longer registered by default.
 
 
 0.15.2 (2015-08-07)
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -17,7 +17,7 @@ From 0.15 to 0.16
       from tchannel import TChannel
 
       @tchannel.register('endpoint', 'json')
-      def endpoint(request, response, proxy):
+      def endpoint(request, response):
           response.write({'resp': 'body'})
 
 
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -31,6 +31,7 @@ From 0.15 to 0.16
       def endpoint(request):
           return Response({'resp': 'body'})
 
+- Stop using `` from tchannel.tornado import TChannel``
 
 - Remove ``retry_delay`` option from ``tchannel.tornado.peer.PeerClientOperation.send``
   method.
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -31,7 +31,7 @@ From 0.15 to 0.16
       def endpoint(request):
           return Response({'resp': 'body'})
 
-- Stop using `` from tchannel.tornado import TChannel``
+- `` from tchannel.tornado import TChannel`` is deprecated.
 
 - Remove ``retry_delay`` option from ``tchannel.tornado.peer.PeerClientOperation.send``
   method.
--- a/examples/benchmark/thrift/server.py
+++ b/examples/benchmark/thrift/server.py
@@ -34,7 +34,7 @@ values = {'hello': 'world'}
 
 
 @app.register(KeyValue)
-def getValue(request, response, tchannel):
+def getValue(request, response):
     key = request.args.key
     value = values.get(key)
 
@@ -45,7 +45,7 @@ def getValue(request, response, tchannel):
 
 
 @app.register(KeyValue)
-def setValue(request, response, tchannel):
+def setValue(request, response):
     key = request.args.key
     value = request.args.value
     values[key] = value
--- a/tchannel/context.py
+++ b/tchannel/context.py
@@ -31,8 +31,8 @@ class TChannelLocal(threading.local):
 _LOCAL = TChannelLocal()
 
 
-class TChannelContext(object):
-    """TChannelContext is used to save necessary context information related
+class RequestContext(object):
+    """RequestContext is used to save necessary context information related
     to current running async thread.
     """
 
--- a/tchannel/context.py
+++ b/tchannel/context.py
@@ -29,16 +29,18 @@ class TChannelLocal(threading.local):
         self.context = None
 
 _LOCAL = TChannelLocal()
-_LOCAL.context = None
 
 
-class Context(object):
-    """Stack context for each TChannel instance."""
+class TChannelContext(object):
+    """TChannelContext is used to save necessary context information related
+    to current running async thread.
+    """
 
     __slots__ = ('parent_tracing', '_old_context',)
 
-    def __init__(self):
-        self.parent_tracing = None
+    def __init__(self, parent_tracing=None):
+        self.parent_tracing = parent_tracing
+        self._old_context = None
 
     def __enter__(self):
         self._old_context = _LOCAL.context
@@ -48,5 +50,9 @@ class Context(object):
         _LOCAL.context = self._old_context
 
 
-def get_local():
-    return _LOCAL
+def get_current_context():
+    """
+
+    :return: stack context in current running aysnc thread.
+    """
+    return _LOCAL.context
--- a/tchannel/context.py
+++ b/tchannel/context.py
@@ -1,0 +1,52 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+import threading
+
+
+class TChannelLocal(threading.local):
+    __slots__ = ('context',)
+
+    def __init__(self):
+        self.context = None
+
+_LOCAL = TChannelLocal()
+_LOCAL.context = None
+
+
+class Context(object):
+    """Stack context for each TChannel instance."""
+
+    __slots__ = ('parent_tracing', '_old_context',)
+
+    def __init__(self):
+        self.parent_tracing = None
+
+    def __enter__(self):
+        self._old_context = _LOCAL.context
+        _LOCAL.context = self
+
+    def __exit__(self, type, value, traceback):
+        _LOCAL.context = self._old_context
+
+
+def get_local():
+    return _LOCAL
--- a/tchannel/context.py
+++ b/tchannel/context.py
@@ -19,7 +19,9 @@
 # THE SOFTWARE.
 
 from __future__ import absolute_import
+
 import threading
+from tornado.stack_context import StackContext
 
 
 class TChannelLocal(threading.local):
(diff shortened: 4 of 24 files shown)


UPGRADE.rst
```diff
@@ -31,6 +31,7 @@ From 0.15 to 0.16
       def endpoint(request):
           return Response({'resp': 'body'})
 
+- Stop using `` from tchannel.tornado import TChannel``
```
abhinav: what

tchannel/context.py
```diff
@@ -0,0 +1,52 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+import threading
+
+
+class TChannelLocal(threading.local):
+    __slots__ = ('context',)
+
```
abhinav: Why is this a class if it's not doing anything?

You can just do,

    _LOCAL = threading.local()
    _LOCAL.context = None

tchannel/context.py
```diff
@@ -0,0 +1,52 @@
...
+class Context(object):
+    """Stack context for each TChannel instance."""
+
+    __slots__ = ('parent_tracing', '_old_context',)
+
+    def __init__(self):
+        self.parent_tracing = None
+
+    def __enter__(self):
+        self._old_context = _LOCAL.context
```
abhinav: If you `__enter__` in a different thread, this will fail because `_LOCAL.context` will not exist. The `context` attribute exists only in the thread that set it. You have to check with `hasattr` first.

tchannel/context.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self):
+        self.context = None
+
+_LOCAL = TChannelLocal()
+_LOCAL.context = None
+
+
+class Context(object):
+    """Stack context for each TChannel instance."""
```
abhinav: Pick a more descriptive name.

Also, this needs a more detailed docstring. The docstring should answer what the purpose of this class is.

tchannel/context.py
```diff
@@ -0,0 +1,52 @@
...
+
+    __slots__ = ('parent_tracing', '_old_context',)
+
+    def __init__(self):
+        self.parent_tracing = None
```
abhinav: Initialize all attributes in `__init__`, including `_old_context`.

tchannel/context.py
```diff
@@ -0,0 +1,52 @@
...
+    def __enter__(self):
+        self._old_context = _LOCAL.context
+        _LOCAL.context = self
+
+    def __exit__(self, type, value, traceback):
+        _LOCAL.context = self._old_context
+
+
+def get_local():
+    return _LOCAL
```
abhinav: Don't expose the actual thread local outside this module.

abhinav: Instead, expose a method that returns the 'current context'

tchannel/context.py
abhinav: Also this isn't really a stack context. I see no mention of stack context anywhere in this module.

tchannel/context.py
junchaowu: so default the context will be part of the thread local.

tchannel/context.py
junchaowu: context will exist. It is part of the thread local class.

tchannel/tchannel.py
```diff
@@ -57,6 +60,15 @@ def __init__(self, name, hostport=None, process_name=None,
...
         self.thrift = schemes.ThriftArgScheme(self)
 
+        self._local = get_local()
+
+    def get_context(self):
+        """
+
+        :return: stack context in current running aysnc thread.
+        """
+        return self._local.context
```
abhinav: No. Don't access the thread local directly.

tchannel/tornado/dispatch.py
```diff
@@ -182,9 +185,11 @@ def handle_call(self, request, connection):
                 )
 
                 # get new top-level resp from controller
-                new_resp = yield gen.maybe_future(
-                    handler.endpoint(new_req)
-                )
+                with StackContext(lambda: _context):
+                    tchannel.get_context().parent_tracing = request.tracing
+                    f = handler.endpoint(new_req)
+
```
abhinav: This should all be encapsulated in that same module that manages the thread. Ideally, this code block should just be

    with Context(parent_tracing):
        # ...

(With a better name than Context, of course.)

tchannel/tornado/dispatch.py
```diff
@@ -287,37 +282,3 @@ def not_found(request, response, proxy):
                 request.service,
             ),
         )
-
-
-class TChannelProxy(object):
-    """TChannel Proxy with additional runtime info
```
abhinav: YASS

tchannel/context.py
junchaowu: TChannel will call this method to get the local. other places will call tchannel.get_context to get context.

tchannel/tornado/dispatch.py
junchaowu: Parent_tracing will be part of context. context != parent_tracing.

tchannel/tornado/peer.py
```diff
@@ -453,6 +453,9 @@ def __init__(self,
         self.tchannel = peer_group.tchannel
         self.service = service
         self.parent_tracing = parent_tracing
+        if self.tchannel.get_context():
+            self.parent_tracing = (parent_tracing or
+                                   self.tchannel.get_context().parent_tracing)
```
abhinav: nit: could be cleaner

    if not self.parent_tracing and self.tchannel.get_context():
        self.parent_tracing = self.tchannel.get_context().parent_tracing

Also, I'm not sure if `get_tracing` needs to exist. what's the use case for people needing to access tho context object?

tchannel/tornado/tchannel.py
```diff
@@ -121,9 +122,6 @@ def __init__(self, name, hostport=None, process_name=None,
         self.event_emitter = EventEmitter()
         self.hooks = EventRegistrar(self.event_emitter)
 
-        from ..zipkin.zipkin_trace import ZipkinTraceHook
-        self.hooks.register(ZipkinTraceHook(tchannel=self))
```
abhinav: Cool so we can drop the hard dependency on thrift?

tests/integration/thrift/test_tornado_client.py
```diff
@@ -82,7 +83,7 @@ def test_thrift_exception(mock_server, thrift_service):
         'thrift',
         method='getItem',
     ).and_raise(thrift_service.ItemDoesNotExist("stahp"))
-
+    mock_server.tchannel.hooks.register(ZipkinTraceHook(tchannel=mock_server))
```
abhinav: If this test isn't testing zipkin integration, why does it need the hook?

junchaowu: The existing test has this mock_trac_record.called assertion. I guess it is testing if trace will be triggered when exception happens in thrift. This is your test. If it doesn't mean that. I will delete it.

tchannel/context.py
abhinav: Ah, if you create it like this it'll never be missing the attribute, even in other threads. Okay.

abhinav: Deleting comment.

tchannel/context.py
abhinav: The thread local and the stack context should not be exposed outside this module. Thread local provides one level of 'local'-ness and the stack context provides the other. These are both details that other parts of the system don't care about. They should be able to do `get_current_context` and get the 'current' `Context` object.

abhinav: And I don't see why `TChannel.get_context` exists either. Users don't need to be able to access the context object. We're managing it transparently for them.

tchannel/tornado/dispatch.py
abhinav: Yes, I know. `parent_tracing` is tracked by context. It is part of the state that it tracks per 'async thread'. I'm assuming `Context` is the class that will hold all this state. Right now, you'll do 

    with Context(parent_tracing):
      # ...

But if we decide to add another piece of state to be tracked, you'd basically change it to

    with Context(parent_tracing, second_state):
        # ...

tests/integration/thrift/test_tornado_client.py
abhinav: Existing tests may have been asserting zipkin behavior because we had it enabled by default previously.

Now that it's opt-in, it might make sense to eliminate these references.
Although, we would still want an integration test of some kind

tchannel/tornado/dispatch.py
breerly: :laughing: :cocktail:

tchannel/tornado/tchannel.py
breerly: Let's do that in another PR

CHANGES.rst
```diff
@@ -19,6 +19,8 @@ Changelog
   the new server API. See the upgrade guide for details.
 - **BREAKING** - remove ``retry_delay`` option in the ``tchannel.tornado.send``
   method.
+- **BREAKING** - remove third arugment ``proxy`` from the server handler interface.
+- **BREAKING** - make the zipkintracehook not default registered.
```
blampe: ``ZipkinTraceHook`` is not longer registered by default.

examples/benchmark/thrift/server.py
```diff
@@ -34,7 +34,7 @@
 
 
 @app.register(KeyValue)
-def getValue(request, response, tchannel):
+def getValue(request, response):
```
breerly: If we're going to update this example let's switch over to `tchannel.TChannel` instead of `tchannel.tornado.TChannel`.

examples/benchmark/thrift/server.py
```diff
@@ -45,7 +45,7 @@ def getValue(request, response, tchannel):
 
 
 @app.register(KeyValue)
```
breerly: Let's call this `tchannel` instead of `app`.

examples/benchmark/thrift/server.py
blampe: +1

tchannel/context.py
```diff
@@ -0,0 +1,58 @@
...
+class TChannelLocal(threading.local):
+    __slots__ = ('context',)
+
+    def __init__(self):
+        self.context = None
+
+_LOCAL = TChannelLocal()
+
+
+class TChannelContext(object):
```
breerly: Let's rename to something more obvious... I think `RequestContext` maybe...

tchannel/testing/vcr/server.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, cassette, unpatch):
         VCRProxy.VCRServiceError,
     ))
     @gen.coroutine
-    def send(self, request, response, channel):
+    def send(self, request, response):
```
breerly: hmmm... so I guess we'll have to get the sync client and vcr working with the top-level `tchannel.TChannel` too

tests/integration/thrift/test_tornado_client.py
```diff
@@ -66,13 +65,7 @@ def test_protocol_error(mock_server, thrift_service):
 
     client = mk_client(thrift_service, mock_server.port, trace=False)
     with pytest.raises(errors.ProtocolError):
-        with patch(
-            'tchannel.zipkin.tracers.TChannelZipkinTracer.record',
```
blampe: Looks like we lost test coverage on the zipkin hooks, can we move these to test cases that intentionally opt-in to zipkin?

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -127,7 +124,7 @@ def test_zipkin_trace(trace_server):
 
 
 @pytest.mark.gen_test
-def test_tcollector_submit(monkeypatch, trace_server):
+def test_tcollector_submit(trace_server):
```
blampe: Don't we need to register the zipkin event hooks?

breerly:
Didn't we want to attach `context` to `tchannel.Request`?

```
@tchannel.json.register
def endpoint(request):
    print request.context    # tchannel.context.RequestContext
```

tests/integration/thrift/test_tornado_client.py
junchaowu: No. We have test_zipkin_trace.py to test zipkin hooks.

tests/integration/trace/test_zipkin_trace.py
junchaowu: No. This unit test is test zipkin trace's submit method. It is not an end to end test. We have 'test_zipkin_trace' test doing end to end test for zipkin. In this test, we register zipkintracehook.

junchaowu:
@breerly whether we want to put context in the request object or whether we take context as argument, need further discussion. At this init stage, I would keep scope of this PR. 

Besides if we decide update all the example or test to use new tchannel.TChannel, that's a large amount code change. Let's put it in a separate PR.

tests/test_context.py
```diff
@@ -0,0 +1,69 @@
...
+import pytest
+from tornado import gen
+
+from tchannel import TChannel
+from tchannel import Response
+from tchannel.context import get_current_context
+
+
+@pytest.mark.gen_test
+def test_context():
```
breerly: `s/test_context/test_context_should_carry_tracing_info/`

breerly: This test should be in `test_tchannel.py` since you're doing an integration test through the public API. This file I would expect to test just the pieces in `context.py`.

tchannel/tornado/dispatch.py
```diff
@@ -182,9 +184,10 @@ def handle_call(self, request, connection):
                 )
 
                 # get new top-level resp from controller
-                new_resp = yield gen.maybe_future(
-                    handler.endpoint(new_req)
-                )
+                with StackContext(lambda: RequestContext(request.tracing)):
+                    f = handler.endpoint(new_req)
```
abhinav: Can we also move the StackContext management to the context module? As said previously, I would like for those details to not be leaked outside that module. This block of code should ideally look like this:

    with RequestContext(request.tracing):
        # ...

Or something similar.

tchannel/context.py
```diff
@@ -0,0 +1,58 @@
...
+        _LOCAL.context = self
+
+    def __exit__(self, type, value, traceback):
+        _LOCAL.context = self._old_context
+
+
+def get_current_context():
+    """
+
+    :return: stack context in current running aysnc thread.
```
abhinav: You're not returning a stack context. You're returning a request context.

tests/test_context.py
```diff
@@ -0,0 +1,73 @@
...
+        scheme=schemes.RAW,
+        service='server',
+        arg1='endpoint1',
+        arg2='req headers',
+        arg3='req body',
+        hostport=server.hostport,
+    )
+
+    assert context[0].parent_tracing.name == 'endpoint1'
+    assert context[1].parent_tracing.name == 'endpoint2'
```
abhinav: Also match the trace ID (since it's part of the same request)

abhinav:
Couple minor issues but the core functionality looks good.

abhinav:
looks good besides comment nit.

PR #125, opened 2015-08-19
bring new api to the sync client

blampe:
WIP

Still need docs and some tests.

* Make sync client inherit from new `TChannel`. 
* Wrap arg schemes and advertise to submit async jobs to the thread loop.
* Remove sync `client_for` (if people really need/want this I think they should be able to use the other `client_for` but I need to confirm).
* `advertise` is returning a `tornado.Response` which I think is wrong.

@breerly @abhinav @junchaowu


breerly:
You'll want call to return a `tchannel.Response`, same for `advertise` although it could be that `tchannel.TChannel` is also busted in this sense.

tchannel/sync/client.py
```diff
@@ -20,18 +20,11 @@
...
-
-from tchannel import tornado as async
-from tchannel.tornado.hyperbahn import FIRST_ADVERTISE_TIME
-from tchannel.tornado.hyperbahn import AdvertiseError
+from tchannel import TChannel
 from threadloop import ThreadLoop
 
 
-class TChannelSyncClient(object):
+class TChannelSyncClient(TChannel):
```
breerly: Since you would never have an async client and sync client in the same application, I think it makes more sense to name this `tchannel.sync.TChannel`. The client code would look like so:

```
from tchannel.sync import TChannel

tchannel = TChannel(...)
```

When we move it out into it's own package, it would just be:

```
from tchannel_sync import TChannel

tchannel = TChannel()
```

From the time you have a tchannel instance on, everything is the same, so the import path is really the only thing that matters.

breerly: Also, that let's us keep the current sync behavior unbroken until we start removing all this old code.

tchannel/sync/client.py
```diff
@@ -52,131 +45,49 @@ class TChannelSyncClient(object):
...
-
-            result = Response(header, body)
-
-            raise gen.Return(result)
+        self.advertise = self._wrap(self.advertise)
 
-        future = self._threadloop.submit(make_request)
+        self.raw = self._wrap(self.raw)
+        self.thrift = self._wrap(self.thrift)
+        self.json = self._wrap(self.json)
```
breerly: Overriding `call` should be enough since all the argscheme's use `call` to ultimately make the request.

blampe: That's what I wanted originally, but the `__call__` method on each scheme is a coroutine, and *that's* the coroutine we need to send over to the threadloop.

```
@coroutine
def __call__():
    response = yield sync_tchannel.call(..) # submit to threadloop

raw(...)  # still returns a tornado future :(
```

tchannel/sync/client.py
blampe: > Also, that let's us keep the current sync behavior unbroken until we start removing all this old code.

Do you mean you want to preserve `TChannelSyncClient` in addition to a `sync.TChannel` to make this backwards compatible?

tchannel/sync/client.py
breerly: That's a very good point :P

breerly:
@abhinav some more eyes here are needed sir

tests/sync/test_client.py
```diff
@@ -63,12 +64,14 @@ def test_advertise_should_result_in_peer_connections(mock_server):
...
-    result = client.advertise(routers)
+    future = client.advertise(routers)
+
+    result = future.result()
 
-    assert result.header == ""
-    # @todo https://github.com/uber/tchannel-python/issues/34
-    assert result.body == json.dumps(body)
-    assert client._async_client.peers.hosts == routers
+    assert result.headers == {'as': 'json'}
```
breerly: well this seems wrong, these are transport headers

breerly:
I think we should rewrite these tests to use real servers

tchannel/sync/client.py
```diff
@@ -20,18 +20,12 @@
...
-from tchannel import tornado as async
-from tchannel import errors
-from tchannel.tornado.hyperbahn import FIRST_ADVERTISE_TIME
 from threadloop import ThreadLoop
 
+from tchannel import TChannel
 
-class TChannelSyncClient(object):
+
+class TChannelSyncClient(TChannel):
```
breerly: Let's rename to:

```
class TChannel(AsyncTChannel):
```

tchannel/sync/client.py
```diff
@@ -52,131 +46,49 @@ class TChannelSyncClient(object):
...
 
-    def __init__(self, name, process_name=None, known_peers=None, trace=False):
+    def __init__(
+        self,
+        name,
+        hostport=None,
+        process_name=None,
+        known_peers=None,
+        trace=False,
+        threadloop=None,
```
breerly: Here the threadloop can be injected but nothing is done with it.

UPGRADE.rst
```diff
@@ -31,7 +31,58 @@ From 0.15 to 0.16
       def endpoint(request):
           return Response({'resp': 'body'})
 
-- `` from tchannel.tornado import TChannel`` is deprecated.
+- ``TChannelSyncClient`` has been replaced with ``tchannel.sync.TChannel``.
+  This new synchronous client has been significantly re-worked to more closely
+  match the asynchronous ``TChannel`` API. ``tchannel.sync.thrift.client_for``
+  has been removed and ``tchannel.thrift.client_for`` should be used instead.
```
abhinav: thrift_request_builder

abhinav:
LGTM besides typo in UPGRADE.rst

tchannel/sync/client.py
blampe: fixed dis

tests/sync/test_client.py
blampe: Figured out that this is because advertise still uses the old client api/response. I'm going to fix that but it will break advertisement with old TChannel instances :/

breerly: Let's just break the top-level tchannel, not `tchannel.tornado.TChannel`

abhinav:
:+1:

PR #129, opened 2015-08-24
update not_found handler for new server API

blampe:
This was failing because sometimes we pass req/res and sometimes we just pass req.

@breerly @abhinav @junchaowu


junchaowu:
lgtm

PR #133, opened 2015-08-24
update mock_server to use new tchannel api

blampe:
* Add `shard_key` to new `Request` object.
* Expose `shard_key` transport header in call args.
* `tests/mock_server` uses new `TChannel` registration/handlers

@junchaowu I had to change the `trace=False` flag to `trace=True`. I don't understand why this is a property of the `TChannel` instance at all. I also don't understand why changing its default value to `True` fixed my tests.

@breerly @abhinav


abhinav:
:+1:

breerly:
Seems like this should be 2 PR's broseph

tchannel/tchannel.py
```diff
@@ -135,6 +146,9 @@ def call(self, scheme, service, arg1, arg2=None, arg3=None,
             transport.SCHEME: scheme,
             transport.CALLER_NAME: self.name,
         }
+        if shard_key:
```
breerly: oh man... this debt is getting out of hand...

we really need to move to allowing a `tchannel.Request` to call soon...

breerly:
:shipit:

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -99,12 +97,15 @@ def test_zipkin_trace(trace_server):
 
     hostport = 'localhost:%d' % trace_server.port
 
-    response = yield tchannel.request(hostport).send(InMemStream(endpoint),
-                                                     InMemStream(hostport),
-                                                     InMemStream(),
-                                                     traceflag=True)
```
junchaowu: because the old way to send request I put traceflag=True in the send call. So it will submit span for tracing.

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -99,12 +97,15 @@ def test_zipkin_trace(trace_server):
...
-    header = yield response.get_header()
-    body = yield response.get_body()
+    response = yield tchannel.raw(
+        service='test-client',
+        hostport=hostport,
+        endpoint=endpoint,
+        headers=hostport,
+    )
```
junchaowu: Set 'trace=True' here. Not to set the default the value to True.

junchaowu:
Need to updated .

breerly:
@blampe get on 'dis shit mayne

blampe:
@junchaowu are you saying that we only submit zipkin traces when we explicitly pass `trace=True`? Is that what we want?

blampe:
@junchaowu have another look, in particular make sure the zipkin test still looks good.

junchaowu:
lgtm

junchaowu:
Look how responsive I am. :)

PR #139, opened 2015-08-26
make client_for compatible with new TChannel

blampe:
@abhinav @junchaowu @breerly

### DIFF
--- a/tchannel/thrift/client.py
+++ b/tchannel/thrift/client.py
@@ -188,6 +188,7 @@ def generate_method(service_module, service_name, method_name):
                 arg2=header,
                 arg3=body,
                 hostport=self.hostport,
+                # TODO: Need to handle these!
                 #headers=self.protocol_headers,
                 #traceflag=self.trace,
             )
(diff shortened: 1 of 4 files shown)


tchannel/thrift/client.py
```diff
@@ -166,17 +167,34 @@ def send(self, *args, **kwargs):
...
+            response = yield self.tchannel.call(
+                scheme=schemes.THRIFT,
+                service=self.service,
+                arg1=endpoint,
+                arg2=header,
+                arg3=body,
+                hostport=self.hostport,
+                #headers=self.protocol_headers,
+                #traceflag=self.trace,
+            )
```
abhinav: Can't you just always call the new API? It currently calls into the old API anyway?

blampe: Some people are still using `tchannel.tornado.TChannel` which doesn't have the new API though.

abhinav:
Ugh, okay. We can remove it when we kill the request.send API.

PR #141, opened 2015-08-27
[vcr] Account for known_peers when making requests

### LINKED ISSUE #138: fun interaction between vcr and sync client
This works:
```python
from tchannel.sync import TChannel
from tchannel import thrift_request_builder
from tchannel.testing import vcr
from tchannel.testing.data.generated.ThriftTest import ThriftTest
                                                                 
service = thrift_request_builder(
    service='thrift-server',
    thrift_module=ThriftTest,
    hostport='localhost:54497'
)
                                                                 
tchannel = TChannel('thrift-client')
                                                           
@vcr.use_cassette('foo.yaml')
def make_request():
                                                                 
    f = tchannel.thrift(
        request=service.testString(thing="req"),
        headers={
            'req': 'header',
        },
    )
                                                                 
    return f.result()

print make_request()
```
Instantiating a `TChannel` after applying `vcr` blocks forever:
```python
@vcr.use_cassette('foo.yaml')
def make_request():
    tchannel = TChannel('thrift-client')
                                                                 
    f = tchannel.thrift(
        request=service.testString(thing="req"),
        headers={
            'req': 'header',
        },
    )
                                                                 
    return f.result()
```

With async `TChannel` it gives a more expected "DummyFuture does not support blocking for results."

cc @abhinav

abhinav:
Previously, if hostPort wasn't specified, the VCR proxy service would end up
making a request back to the peer which made the request and that would fail
because that peer doesn't know how to respond to that request. The fix
involves tracking the known peers of a TChannel at the time the request was
made and sending it to the VCR proxy service.

Actually resolves #138.

### DIFF
--- a/tchannel/testing/vcr/patch.py
+++ b/tchannel/testing/vcr/patch.py
@@ -53,6 +53,7 @@ class PatchedClientOperation(object):
     def __init__(
         self,
         vcr_client,
+        original_tchannel,
         hostport=None,
         service=None,
         arg_scheme=None,
@@ -64,6 +65,7 @@ class PatchedClientOperation(object):
         self.hostport = hostport
         self.service = service or ''
         self.arg_scheme = arg_scheme or schemes.DEFAULT
+        self.original_tchannel = original_tchannel
 
         # TODO what to do with retry, parent_tracing and score_threshold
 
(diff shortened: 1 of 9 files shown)


tchannel/testing/vcr/patch.py
```diff
@@ -83,6 +85,7 @@ def send(self, arg1, arg2, arg3,
         vcr_request = VCRProxy.Request(
             serviceName=self.service.encode('utf-8'),
             hostPort=self.hostport,
+            knownPeers=self.original_tchannel.peers.hosts,
```
blampe: I think this assumes its getting a `tchannel.tornado.TChannel`, can we make it also work with `tchannel.TChannel`?

abhinav: This logic patches the tornado.TChannel object. It's guaranteed to get that.

This logic will probably be changed when we invert the flow so that `request.send` is implemented on top of `call`.

blampe:
:+1:

PR #147, opened 2015-09-02
accept host file in advertise

ShanniLi:
r @junchaowu @blampe


tchannel/tchannel.py
```diff
@@ -206,7 +206,7 @@ def decorator(fn):
             return decorator(handler)
 
     @gen.coroutine
-    def advertise(self, routers, name=None, timeout=None):
+    def advertise(self, routers, name=None, timeout=None, router_file=None):
```
junchaowu: Need to explain what the router_file's format. Other users can put any format content in the file.

tchannel/tchannel.py
```diff
@@ -240,6 +243,12 @@ def advertise(self, routers, name=None, timeout=None):
             When unable to make our first advertise request to Hyperbahn.
             Subsequent requests may fail but will be ignored.
         """
+        if routers is None and router_file is not None:
+            # should just let the exceptions fly
+            with open(router_file, 'r') as json_data:
+                routers = json.load(json_data)
+                json_data.close()
```
junchaowu: no need to call close since it's using with

tchannel/tchannel.py
abhinav: You need to give `routers` a default value as well.

The way to use this API should not be `tchannel.advertise(None, router_file="foo")`.

tchannel/tchannel.py
junchaowu: Add exception catch in the json load

abhinav: What about the case when both routers and router_file are None?

tchannel/tchannel.py
```diff
@@ -232,6 +232,9 @@ def advertise(self, routers, name=None, timeout=None):
             The timeout (in seconds) for the initial advertise attempt.
             Defaults to 30 seconds.
 
+        :param router_file:
+            The host file that contains the routers information.
+
```
abhinav: Documentation should mention that exactly one of routers and router_file must be provided.

tchannel/tornado/hyperbahn.py
```diff
@@ -101,12 +101,17 @@ def _advertise_with_backoff(tchannel, service, timeout=None):
...
     """Advertise with Hyperbahn.
 
     See :py:class:`tchannel.TChannel.advertise`.
     """
     timeout = timeout or FIRST_ADVERTISE_TIME
+    if routers is None and router_file is not None:
+        # should just let the exceptions fly
+        with open(router_file, 'r') as json_data:
+            routers = json.load(json_data)
+            json_data.close()
```
abhinav: Same comments as above apply

tests/tornado/test_hyperbahn.py
```diff
@@ -44,6 +46,31 @@ def test_new_client_establishes_peers():
...
+        channel,
+        'baz',
+        None,
+        None,
+        host_path
+    )
+
+    with open(host_path, 'r') as json_data:
+        routers = json.load(json_data)
+        json_data.close()
```
junchaowu: same here.

tests/tornado/test_hyperbahn.py
```diff
@@ -44,6 +46,31 @@ def test_new_client_establishes_peers():
         assert channel.peers.lookup(router)
 
 
+def test_new_client_establishes_peers_from_file():
+
+    host_path = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)),
+        '../data/hosts.json',
+    )
```
abhinav: For the tests, it's safe to assume that the current directory is the project root.

tests/tornado/test_hyperbahn.py
```diff
@@ -44,6 +46,31 @@ def test_new_client_establishes_peers():
...
+    # TChannel knows about one of the peers already.
+    channel = TChannel('test', known_peers=['127.0.0.1:23002'])
+
+    hyperbahn.advertise(
+        channel,
+        'baz',
+        None,
+        None,
+        host_path
+    )
```
abhinav: You probably want to use `channel.advertise`

abhinav: Also, advertise is an async operation. You need to `yield` it.

tchannel/tchannel.py
abhinav: @junchaowu It's probably okay to let the `json.load` exception bubble up to the user without us catching it. Any reason to catch it?

tests/tornado/test_tchannel.py
```diff
@@ -88,3 +89,23 @@ def test_advertise_should_listen_if_not_called_yet(tchannel, mock_server):
     )
 
     assert tchannel.is_listening() is True
+
+
+def test_advertise_should_take_a_router_file(tchannel, mock_server):
+
+    mock_server.expect_call('ad', 'json').and_write(
```
blampe: mock_server is kinda lame and we should avoid introducing new uses. Check out `tests/test_tchannel.py` to see how to write a server inline.

tests/tornado/test_tchannel.py
```diff
@@ -88,3 +89,23 @@ def test_advertise_should_listen_if_not_called_yet(tchannel, mock_server):
...
+        headers="",
+        body={"hello": "world"},
+    )
+
+    host_path = os.path.join(
```
blampe: There's a bug here because this file points to hosts listening on 21300 but the server you're actually running is listening on a random port. You could probably forego `hosts.json` and just write the server's hostport to a temporary file (https://pytest.org/latest/tmpdir.html).

blampe:
Love the initiative :)

You should add a note to the change log.

tchannel/tchannel.py
junchaowu: Log the exception and then throw it.

tchannel/tchannel.py
ShanniLi: all fixed.

tchannel/tchannel.py
ShanniLi: agree with @abhinav. I don't really see the reason to do anything about the exception here.
In general, it can be two reasons: 1) programmer error: the developer should fix their code. 2) host.json file gets corrupted: the program should stop right away, i.e., what's the point to continue if there is no way connect to Hyperbahn?

tests/tornado/test_hyperbahn.py
ShanniLi: @abhinav, 1) I meant to test hyperbahn.advertise. 2) Will using yield make any difference in this test.

tchannel/tchannel.py
junchaowu: If we can log the issue , it will go to 'tchannel' log and will be easy for us the debug why it crashes.

junchaowu: with open(filepath, 'r') as f:
try:
 routers = ujson.load(f)
 except (IOError, OSError, ValueError):
     logger.exception('Failed to read seed routers list.')
     raise

tests/tornado/test_hyperbahn.py
ShanniLi: @abhinav Good to know. I will keep it this way since it doesn't hurt. :)

tests/tornado/test_hyperbahn.py
abhinav: Yielding advertise fails if the first advertise fails, but passes and resumes after that. So you do want to test for that, because otherwise you're not really testing if advertise succeeds or fails.

ShanniLi: tried yield. This test itself doesn't support it.

abhinav: Did you add `@pytest.mark.gen_test`?

tests/tornado/test_tchannel.py
ShanniLi: Reverted this change, since I tested a wrong tchannel objected.
The new change is in tests/test_tchannel.py

ShanniLi:
All comments addressed.

CHANGES.rst
```diff
@@ -198,3 +198,4 @@ Changes by Version
 ------------------
 
 - Initial release.
+- Accept host file in advertise
```
blampe: Goes under 0.16.2 and please add some more usage detail.

tchannel/tchannel.py
```diff
@@ -232,6 +236,12 @@ def advertise(self, routers, name=None, timeout=None):
             The timeout (in seconds) for the initial advertise attempt.
             Defaults to 30 seconds.
 
+        :param router_file:
+            The host file that contains the routers information. The file
+            should contain a JSON stringified format of the routers parameter.
+            Either routers or router_file should be provided. If both provided,
```
blampe: ``routers`` or ``router_file``

tchannel/tchannel.py
```diff
@@ -240,6 +250,15 @@ def advertise(self, routers, name=None, timeout=None):
             When unable to make our first advertise request to Hyperbahn.
             Subsequent requests may fail but will be ignored.
         """
+        if routers is None and router_file is not None:
+            # should just let the exceptions fly
+            with open(router_file, 'r') as json_data:
+                try:
```
blampe: try/except need to wrap the with block in case we can't open the file.

tchannel/tornado/hyperbahn.py
```diff
@@ -101,12 +101,20 @@ def _advertise_with_backoff(tchannel, service, timeout=None):
...
 
     See :py:class:`tchannel.TChannel.advertise`.
     """
     timeout = timeout or FIRST_ADVERTISE_TIME
+    if routers is None and router_file is not None:
+        # should just let the exceptions fly
+        with open(router_file, 'r') as json_data:
+            try:
+                routers = json.load(json_data)
+            except (IOError, OSError, ValueError):
```
blampe: Same, probably worth a test case to try opening a non-existent file.

Actually @breerly how would you feel about only implementing this in the new TChannel API?

tests/test_tchannel.py
```diff
@@ -278,3 +281,33 @@ def test_event_hook_register():
...
+        patch(
+            'tchannel.tornado.TChannel.advertise',
+            autospec=True,
+        )
+    ) as mock_advertise:
+        f = gen.Future()
+        mock_advertise.return_value = f
+        f.set_result(Response())
+        tchannel.advertise(
+            None,
```
blampe: `tchannel.advertise(router_file=host_path)`

tchannel/tchannel.py
ShanniLi: router_file

tests/tornado/test_hyperbahn.py
ShanniLi: Yes, I missed @pytest.mark.gen_test
On the other hand, the purpose of this test is to make sure that hyperbahn.advertise can consume the host file. Use yield will make the test fail since the ad itself will fail. Not what I want.

tchannel/tchannel.py
```diff
@@ -232,6 +236,12 @@ def advertise(self, routers, name=None, timeout=None):
             The timeout (in seconds) for the initial advertise attempt.
             Defaults to 30 seconds.
 
+        :param router_file:
+            The host file that contains the routers information. The file
+            should contain a JSON stringified format of the routers parameter.
+            Either routers or router_file should be provided. If both provided,
+            routers will be used.
```
abhinav: I would prefer an exception being raised instead of silently using `routers` since passing both is technically invalid usage.

tchannel/tornado/hyperbahn.py
```diff
@@ -101,12 +101,20 @@ def _advertise_with_backoff(tchannel, service, timeout=None):
 
 
 @tornado.gen.coroutine
-def advertise(tchannel, service, routers, timeout=None):
+def advertise(tchannel, service, routers, timeout=None, router_file=None):
```
abhinav: `routers=None`?

tchannel/tchannel.py
ShanniLi: I am OK either way. I cannot find an example in the code about parameter validation. Can you provide an example?

abhinav: We haven't yet defined an exception specific to invalid parameters. Just check if they're both None or if they're both not None and raise a `ValueError` with a message like "Only one of routers and router_file must be provided."

I'm fairly sure we just have assertions in some part of the code, but generally a bad idea for public APIs.

tchannel/tchannel.py
breerly: wondering if we should just overload the variable... if it looks like a file then treat it as a file

junchaowu:
lgtm

breerly:
@ShanniLi I'd appreciate you addressing all comments before merging next time.

tchannel/tchannel.py
ShanniLi: Merging two semantics into one variable seems to be an anti-pattern to me. It is confusing to consumers.

On the other hand, node implementation has two separate parameters as well. It is good to keep xlang interfaces consistent.

ShanniLi:
@breerly oh, sorry, missed yours. :( I have replied to your comment.

tchannel/tornado/tchannel.py
```diff
@@ -135,7 +135,7 @@ def trace(self):
         else:
             return self._trace
 
-    def advertise(self, routers, name=None, timeout=None):
+    def advertise(self, routers, name=None, timeout=None, router_file=None):
```
abhinav: this needs `routers=None` too

PR #148, opened 2015-09-02
Update email address

abhinav:
Changing address temporarily until we have a public address for the project.

CC @jwolski


jwolski:
lgtm

PR #157, opened 2015-09-08
expose backlog and set nodelay

blampe:
@abhinav @breerly @junchaowu

### DIFF
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -418,6 +418,7 @@ class TornadoConnection(object):
         stream = tornado.iostream.IOStream(
             socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         )
+        stream.set_nodelay(True)
 
         log.debug("Connecting to %s", hostport)
         try:
--- a/tchannel/tornado/tchannel.py
+++ b/tchannel/tornado/tchannel.py
@@ -135,7 +135,14 @@ class TChannel(object):
         else:
             return self._trace
 
-    def advertise(self, routers, name=None, timeout=None, router_file=None, backlog=2048):
+    def advertise(
+        self,
+        routers,
+        name=None,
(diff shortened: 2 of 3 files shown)


tchannel/tornado/connection.py
```diff
@@ -418,6 +418,7 @@ def outgoing(cls, hostport, process_name=None, serve_hostport=None,
         stream = tornado.iostream.IOStream(
             socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         )
+        stream.set_nodelay(True)
```
abhinav: Why?

tchannel/tornado/tchannel.py
```diff
@@ -221,7 +228,7 @@ def request(self,
                                   retry=retry,
                                   **kwargs)
 
-    def listen(self, port=None):
+    def listen(self, port=None, backlog=2048):
```
abhinav: Use `None` for default in parameter list and set default value inside the body. This allows you to not duplicate the default value in all the callers (since they can just pass None).

abhinav:
Can you explain why the `set_nodelay` is required?

tchannel/tornado/connection.py
blampe: http://tornado.readthedocs.org/en/latest/iostream.html#tornado.iostream.BaseIOStream.set_nodelay

https://en.wikipedia.org/wiki/Nagle%27s_algorithm#Interactions_with_real-time_systems

abhinav:
:ok:

tchannel/tornado/connection.py
breerly: Can you add a comment above this line so we know why we did this in the future?

junchaowu: interesting one

junchaowu:
:ok:

PR #160, opened 2015-09-08
make listen thread safe.

junchaowu:
@blampe @breerly @abhinav

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,8 +5,7 @@ Changes by Version
 0.16.3 (unreleased)
 -------------------
 
-- Fix ``TChannel.listen`` to be thread safe. It won't allow to listen on
-  different port if instance is already listening.
+- Make ``TChannel.listen`` thread-safe and idempotent.
 
 
 0.16.2 (2015-09-04)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,13 @@ Changes by Version
 ==================
 
 
+0.16.3 (unreleased)
+-------------------
+
+- Fix ``TChannel.listen`` to be thread safe. It won't allow to listen on
+  different port if instance is already listening.
+
+
 0.16.2 (2015-09-04)
 -------------------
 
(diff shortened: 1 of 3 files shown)


tchannel/tchannel.py
```diff
@@ -176,6 +177,16 @@ def call(self, scheme, service, arg1, arg2=None, arg3=None,
         raise gen.Return(result)
 
     def listen(self, port=None):
+        if self._dep_tchannel.is_listening():
+            listening_port = int(self._dep_tchannel.hostport.rsplit(":")[1])
+            if port and port != listening_port:
```
abhinav: So this doesn't make it fully thread-safe. The implementation of `dep_tchannel.listen` is:

    if self.is_listening: raise
    # start listening
    self._server = server  # < this is what sets is_listening to True

So there's this huge area in the middle where you could end up listening twice. You want to acquire a lock or something the first thing when listen() is called so that if another listen has to wait for that to finish before it can check is_listening.

CHANGES.rst
```diff
@@ -2,6 +2,13 @@ Changes by Version
 ==================
 
 
+0.16.3 (unreleased)
+-------------------
+
+- Fix ``TChannel.listen`` to be thread safe. It won't allow to listen on
+  different port if instance is already listening.
```
abhinav: Make ``TChannel.listen`` thread-safe and idempotent.

tchannel/tchannel.py
```diff
@@ -176,7 +179,18 @@ def call(self, scheme, service, arg1, arg2=None, arg3=None,
         raise gen.Return(result)
 
     def listen(self, port=None):
-        return self._dep_tchannel.listen(port)
+        with self._listen_lock:
+            if self._dep_tchannel.is_listening():
+                listening_port = int(self._dep_tchannel.hostport.rsplit(":")[1])
```
abhinav: The number of times we're doing this--we should have a `port` attribute exposed. Let's do that in a separate PR later.

abhinav:
LGTM. @blampe, @breerly?

blampe:
:+1:

PR #162, opened 2015-09-09
Fix status code being unset for Thrift using the new server-side API.

abhinav:


abhinav:
@blampe @breerly @junchaowu

breerly:
:it:

PR #163, opened 2015-09-10
fix ttl not set bug

junchaowu:
@blampe @abhinav


abhinav:
Piggy backing status code fix on this PR

abhinav:
:ok: once the test passes

PR #166, opened 2015-09-10
update the example of sync client in the comments to use the right timeout unit.

junchaowu:
@blampe


blampe:
Ok if tests pass. Why is our unit seconds?? :(

junchaowu:
This just changes the comments. The code has been merge Yesterday.....
Why is second because it is a convention in most of python timing method the unit is second. That's why I use sec as basically unit.

PR #168, opened 2015-09-10
tcollector thrift file update

junchaowu:


blampe:
Integration with the new API will come later?

junchaowu:
If you mean ThriftRW, at least more integration work need to be done in abg's pr

PR #170, opened 2015-09-11
add Meta health endpoint

junchaowu:
@blampe @ShanniLi @Raynos


tests/test_health.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def user_health(request):
+    return HealthStatus(ok=False, message="from me")
+
+
+@pytest.mark.gen_test
+def test_user_health():
+    server = TChannel("health_test_server")
+    server.register_health_handler(user_health, method='health')
```
blampe: Love the test! Seeing how this looks to end users is really helpful.

We shouldn't have a separate registration flow for health. It should be the same as any other endpoint:

```
@tchannel.thrift.register(Meta)
def health(request):
    return HealthStatus(ok=True)
```

We already register the "not found" handler by default, this could go there: https://github.com/uber/tchannel-python/blob/master/tchannel/tornado/dispatch.py#L67

tests/test_health.py
```diff
@@ -0,0 +1,64 @@
...
+    resp = yield client.thrift(request=service.health())
+    assert resp.body.ok is True
+    assert resp.body.message is None
+
+
+@pytest.mark.gen_test
+def test_user_health():
+    server = TChannel("health_test_server")
+
+    @server.thrift.register(Meta, method="health")
```
blampe: Would still work if it was `def health(` right?

blampe:
lgtm but need to save this for 0.17

tchannel/health/health.py
```diff
@@ -0,0 +1,27 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+
+from .thrift.ttypes import HealthStatus
+
+
+def health(request):
+    return HealthStatus(ok=True)
```
ShanniLi: In node, it allows registering a callback from service to return a health status. This just returns status true?

ShanniLi:
why don't use thriftrw?

ShanniLi:
Some questions. Other than that, lgtm.

tchannel/health/health.py
blampe: This is the default healtcheck, you can override as in the "test_user_health" test.

tests/test_health.py
junchaowu: Yea totally. Here I just use a different example that the method name doesn't have to be same.

junchaowu:
@ShanniLi  Our python thriftrw hasn't been integrated into TChannel yet. When we worked thriftrw integration, we will update all places.

PR #171, opened 2015-09-14
add docs around request context

blampe:
@junchaowu 

I think docs were broken due to the warning in `UPGRADE.rst`.


blampe:
cc @breerly since I think you're the only one around today :)

breerly:
:shipit:

PR #179, opened 2015-09-16
remove generated code for the health endpoint

blampe:
I think we have a problem with `thrift.load` requiring a service name.

This makes sense for callers because they can pass in either the hostport or the name to use on Hyperbahn.

For servers it doesn't make sense. You already initialized your TChannel instance with your name/hostport, and you're just using `thrift.load` to register your handlers. The `Meta::health` endpoint here is a good example -- I pass in a service of `None` and it's totally fine.

@breerly @junchaowu


tests/test_health.py
```diff
@@ -33,12 +33,14 @@ def test_default_health():
     server.listen()
 
     client = TChannel("health_test_client")
-    service = thrift_request_builder(
-        service='meta',
-        thrift_module=Meta,
+
+    service = thrift.load(
+        'health_test_server',
```
junchaowu: little bit weird to set service name in the thrift.load.

junchaowu:
others look good.

tests/test_health.py
blampe: This one makes sense because it's a downstream call and this is the name hyperbahn would get.

CHANGES.rst
```diff
@@ -4,7 +4,8 @@ Changes by Version
 0.17.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Made "service" optional for ``thrift.load()``. The first argument should be a
+  path, but backwards compatibility is provided for 0.17.0.
```
breerly: thanks :)

tchannel/thrift/rw.py
```diff
@@ -37,7 +37,7 @@
 from .module import ThriftRequest
 
 
-def load(service, path, hostport=None, module_name=None):
+def load(path, service=None, hostport=None, module_name=None):
```
breerly: cool :)

tests/schemes/test_thrift.py
```diff
@@ -134,8 +130,8 @@ def service(server, use_thriftrw_client):
 
     if use_thriftrw_client:
         return thrift.load(
-            'server',
             'tests/data/idls/ThriftTest.thrift',
+            'server',
```
breerly: you should switch these to kwargs

tests/test_health.py
breerly: agreed w @blampe here

breerly: lets use kwargs whenever calling thrift.load

breerly: also we want the service name to be second

PR #180, opened 2015-09-16
WIP graceful shut down tchannel server

junchaowu:
As I work on this task, I realize it will become fairly large code change. Therefore I share the WIP PR earlier.

Here are code parts I will touch and modified.
- [ ] Add tchannel ``stop`` api
- [ ] Move the some states from message_factory to connection class. The refactor for the message_factory can be continued. Considering the scope, I only do the parts that related to drain portion.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.17.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Add ``TChannel.stop()`` method to gracefully shutdown TChannel server.
 
 
 0.17.2 (2015-09-18)
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -26,7 +26,6 @@ import json
 import logging
 
 from threading import Lock
-from tchannel.tornado.tchannel import State
 from tornado import gen
 
 from . import schemes
@@ -39,6 +38,7 @@ from .health import health
 from .health import Meta
 from .response import Response, TransportHeaders
 from .tornado import TChannel as DeprecatedTChannel
+from .tornado.tchannel import State
 from .tornado.dispatch import RequestDispatcher as DeprecatedDispatcher
 
 log = logging.getLogger('tchannel')
@@ -324,7 +324,8 @@ class TChannel(object):
 
     @gen.coroutine
     def stop(self, reason=None):
-        """Gracefully shut down tchannel.
+        """Gracefully shut down TChannel instance.
+
         :param reason:
             User can specify the reason for the stop action.
         """
@@ -354,11 +355,9 @@ class TChannel(object):
             draining.
 
         .. code:: python
-            def exempt_sample(service):
-                if service == 'hello':
-                    return True
 
-                return False
+            def exempt_sample(service):
+                return service == 'hello'
 
         """
 
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -27,6 +27,7 @@ import logging
 
 from threading import Lock
 from tornado import gen
+from tornado import ioloop
 
 from . import schemes
 from . import transport
@@ -319,3 +320,7 @@ class TChannel(object):
         response = Response(json.loads(body), headers or {})
 
         raise gen.Return(response)
+
+    def stop(self, reason=None):
+        return self._dep_tchannel.stop(reason)
+
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -344,6 +344,7 @@ class TChannel(object):
     @gen.coroutine
     def drain(self, reason=None, exempt=None):
         """Drain the existing connections, and stop taking new requests.
+
         :param reason:
             User can specify the reason for the drain action.
         :param exempt:
@@ -360,7 +361,9 @@ class TChannel(object):
                 return False
 
         """
+
         self._dep_tchannel._server.drain()
+
         yield [
             peer.drain(reason, exempt) for peer in
             self._dep_tchannel.peer_group.peers
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -26,6 +26,7 @@ import json
 import logging
 
 from threading import Lock
+from tchannel.tornado.tchannel import State
 from tornado import gen
 
 from . import schemes
@@ -109,6 +110,7 @@ class TChannel(object):
         )
 
         self.name = name
+        self.known_peers = known_peers
 
         # set arg schemes
         self.raw = schemes.RawArgScheme(self)
@@ -320,8 +322,46 @@ class TChannel(object):
 
         raise gen.Return(response)
 
+    @gen.coroutine
     def stop(self, reason=None):
-        self._dep_tchannel.stop(reason)
+        """Gracefully shut down tchannel.
+        :param reason:
+            User can specify the reason for the stop action.
+        """
+        yield self.drain(reason)
+        self._dep_tchannel.peer_group.clear()
+        self._dep_tchannel._server.stop()
+        self._dep_tchannel._server = None
+        self._dep_tchannel._state = State.ready
+        if self.known_peers:
+            for peer_hostport in self.known_peers:
+                self._dep_tchannel.peer_group.add(peer_hostport)
 
+        if self._dep_tchannel.advertise_loop:
+            self._dep_tchannel.advertise_loop.stop()
+            self._dep_tchannel.advertise_loop = None
+
+    @gen.coroutine
     def drain(self, reason=None, exempt=None):
-        return self._dep_tchannel.drain(reason, exempt)
+        """Drain the existing connections, and stop taking new requests.
+        :param reason:
+            User can specify the reason for the drain action.
+        :param exempt:
+            exempt should be a function that takes service name as string
+            argument and return bool. It provides a way for use to specify what
+            service requests that will continue to be processed during
+            draining.
+
+        .. code:: python
+            def exempt_sample(service):
+                if service == 'hello':
+                    return True
+
+                return False
+
+        """
+        self._dep_tchannel._server.drain()
+        yield [
+            peer.drain(reason, exempt) for peer in
+            self._dep_tchannel.peer_group.peers
+        ]
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -343,25 +343,14 @@ class TChannel(object):
             self._dep_tchannel.advertise_loop = None
 
     @gen.coroutine
-    def drain(self, reason=None, exempt=None):
+    def drain(self, reason=None):
         """Drain the existing connections, and stop taking new requests.
 
         :param reason:
             User can specify the reason for the drain action.
-        :param exempt:
-            exempt should be a function that takes service name as string
-            argument and return bool. It provides a way for use to specify what
-            service requests that will continue to be processed during
-            draining.
-
-        .. code:: python
-
-            def exempt_sample(service):
-                return service == 'hello'
-
         """
 
         yield [
-            peer.drain(reason, exempt) for peer in
+            peer.drain(reason) for peer in
             self._dep_tchannel.peer_group.peers
         ]
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -361,8 +361,6 @@ class TChannel(object):
 
         """
 
-        self._dep_tchannel._server.drain()
-
         yield [
             peer.drain(reason, exempt) for peer in
             self._dep_tchannel.peer_group.peers
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -27,7 +27,6 @@ import logging
 
 from threading import Lock
 from tornado import gen
-from tornado import ioloop
 
 from . import schemes
 from . import transport
@@ -322,5 +321,7 @@ class TChannel(object):
         raise gen.Return(response)
 
     def stop(self, reason=None):
-        return self._dep_tchannel.stop(reason)
+        self._dep_tchannel.stop(reason)
 
+    def drain(self, reason=None, exempt=None):
+        return self._dep_tchannel.drain(reason, exempt)
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -24,16 +24,18 @@ import logging
 import os
 import socket
 import sys
-from tornado.concurrent import Future
 
 import tornado.gen
 import tornado.iostream
+from tornado.concurrent import Future
+
 from .. import errors
 from .. import frame
 from .. import glossary
 from .. import messages
-from ..errors import NetworkError, DeclinedError
+from ..errors import DeclinedError
 from ..errors import FatalProtocolError
+from ..errors import NetworkError
 from ..errors import TChannelError
 from ..event import EventType
 from ..io import BytesIO
@@ -53,6 +55,26 @@ except ImportError:
     from Queue import Empty as QueueEmpty
 
 log = logging.getLogger('tchannel')
+
+
+class _Drain(object):
+
+    DEFAULT_REASON = "Server has stopped accepting new requests."
+
+    def __init__(self, reason=None, exempt=None, future=None):
+        """
+        :param reason:
+            User can specify the reason for the drain action.
+        :param exempt:
+            exempt function to exempt request for particular service for
+            draining.
+        :param future:
+            The future is used to tell when all in progress requests have been
+            processed. In other words, draining is completed.
+        """
+        self.future = future or Future()
+        self.reason = reason or self.DEFAULT_REASON
+        self.exempt = exempt
 
 
 class TornadoConnection(object):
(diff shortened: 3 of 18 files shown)


tchannel/tchannel.py
```diff
@@ -322,3 +322,9 @@ def advertise(self, routers=None, name=None, timeout=None,
         response = Response(json.loads(body), headers or {})
 
         raise gen.Return(response)
+
+    def stop(self, reason=None):
```
breerly: should this be public?

tchannel/tchannel.py
```diff
@@ -322,3 +322,9 @@ def advertise(self, routers=None, name=None, timeout=None,
...
+        self._dep_tchannel.stop(reason)
+
+    def drain(self, reason=None, exempt=None):
```
breerly: should this be public?

tchannel/tchannel.py
```diff
@@ -322,3 +322,9 @@ def advertise(self, routers=None, name=None, timeout=None,
         response = Response(json.loads(body), headers or {})
 
         raise gen.Return(response)
+
+    def stop(self, reason=None):
+        self._dep_tchannel.stop(reason)
```
blampe: We need to stop building things into the old API. We can build *around* it, but we shouldn't also be adding features to the old API.

tchannel/tchannel.py
blampe: Docs. What are reason and exempt and why would I need to pass them?

tchannel/tornado/connection.py
```diff
@@ -24,15 +24,15 @@
...
 
 import tornado.gen
 import tornado.iostream
-
 from .. import errors
 from .. import frame
 from .. import glossary
 from .. import messages
-from ..errors import NetworkError
+from ..errors import NetworkError, DeclinedError
```
blampe: One import per line.

tchannel/tornado/connection.py
```diff
@@ -79,6 +79,8 @@ class TornadoConnection(object):
     CALL_REQ_TYPES = frozenset([Types.CALL_REQ, Types.CALL_REQ_CONTINUE])
     CALL_RES_TYPES = frozenset([Types.CALL_RES, Types.CALL_RES_CONTINUE])
 
+    DEFAULT_REASON = "Server stops accepting new requests."
```
blampe: has stopped

blampe: Although it seems like this is something we shouldn't care about.

tchannel/tornado/connection.py
```diff
@@ -121,7 +120,13 @@ def __init__(self, connection, tchannel=None):
         self._loop_running = False
 
         self.tchannel = tchannel
+        self.incoming_requests = {}
+        self.incoming_responses = {}
```
blampe: What's an incoming response? I think should be "inflight"?

tchannel/tornado/connection.py
```diff
@@ -121,7 +120,13 @@ def __init__(self, connection, tchannel=None):
...
 
+        self.drain_future = None
+        self.draining = False
```
blampe: Lots of state here, smells like you need an object.

tchannel/tornado/connection.py
```diff
@@ -192,6 +195,8 @@ def on_error(future):
 
             if isinstance(exception, tornado.iostream.StreamClosedError):
                 self.close()
+            if message_future.running():
```
blampe: Comment? What case is this covering? Is this tested?

tchannel/tornado/connection.py
junchaowu: It is place to save the "inflight" incoming requests, but considering we may add something for outgoing req/resp in the future, I'd rather name it incoming

tchannel/tornado/connection.py
```diff
@@ -205,50 +210,64 @@ def _loop(self):
         #
         # Must be started only after the handshake has been performed.
         self._loop_running = True
+
         while not self.closed:
```
blampe: This whole loop was already complicated and hard to refactor and this is just making it worse.

tchannel/tornado/connection.py
```diff
@@ -452,12 +466,26 @@ def serve(self, handler):
         """
         assert handler, "handler is required"
         assert self._loop_running, "Finish the handshake first"
-
         while not self.closed:
             message = yield self.await()
-
             try:
-                handler(message, self)
+                if self.draining:
```
blampe: Look at how many levels of indentation and branching there are! Something is wrong.

When you're shutting down why not just remove all of your handlers and replace your fallback handler with one that returns "declined"?

tchannel/tornado/dispatch.py
```diff
@@ -85,6 +84,7 @@ def handle(self, message, connection):
         handler_name = "handle_" + self._HANDLER_NAMES[message.message_type]
         return getattr(self, handler_name)(message, connection)
 
+    @gen.coroutine
```
blampe: No!

tchannel/tchannel.py
junchaowu: removed the stop and drain in the tornado/tchannel

tchannel/tornado/connection.py
junchaowu: Lots of our tests actually cover this case. For example a simple client & server case, client sends one request to server, gets response and client exists. At the end, this connection is closed however the server side's , it ``yield _recv()`` and this message_future never got triggered.

junchaowu: This is a blank spot  for io_loop. It won't throw exceptions for those never triggered ``yield futures`` during shut down process.

tchannel/tornado/dispatch.py
junchaowu: removed

tchannel/tornado/connection.py
junchaowu: You can't just remove all handlers. Think about the purpose.
1. there is exempt function exist, which mean some of the service will still be accepted during draining.
2. It needs to be able to relaunch after stop to solve the global test instance issue. If I remove all handlers, there is no way it can be relaunch

tchannel/tornado/connection.py
blampe: I don't understand what we would put in the 'outgoing_request' map and why that isn't covered by considering it in-flight.

What do these maps actually represent? It's not clear from the code. What's the key/value type? Are they fragmented messages we're building? Are they requests waiting for responses?

tchannel/tornado/connection.py
blampe: Tests only cover behavior if they make assertions about that behavior. Lots of tests might be hitting this code path, but this is behavior we need to explicitly test. Are we doing that?

tchannel/tchannel.py
```diff
@@ -26,6 +26,7 @@
 import logging
 
 from threading import Lock
+from tchannel.tornado.tchannel import State
```
blampe: We should not be adding any new code to `tchannel.tornado.tchannel`.

tchannel/tchannel.py
```diff
@@ -319,3 +321,50 @@ def advertise(self, routers=None, name=None, timeout=None,
         response = Response(json.loads(body), headers or {})
 
         raise gen.Return(response)
+
+    @gen.coroutine
+    def stop(self, reason=None):
```
blampe: Eventually this will need to call unadvertise right?

tchannel/tchannel.py
junchaowu: this is existing one in the old tchannel.

tchannel/tchannel.py
```diff
@@ -319,3 +321,50 @@ def advertise(self, routers=None, name=None, timeout=None,
...
+            User can specify the reason for the stop action.
+        """
+        yield self.drain(reason)
+        self._dep_tchannel.peer_group.clear()
+        self._dep_tchannel._server.stop()
+        self._dep_tchannel._server = None
+        self._dep_tchannel._state = State.ready
+        if self.known_peers:
+            for peer_hostport in self.known_peers:
+                self._dep_tchannel.peer_group.add(peer_hostport)
```
blampe: Why are we adding peers when we stop?

tchannel/tchannel.py
```diff
@@ -319,3 +321,50 @@ def advertise(self, routers=None, name=None, timeout=None,
...
+            User can specify the reason for the drain action.
+        :param exempt:
+            exempt should be a function that takes service name as string
+            argument and return bool. It provides a way for use to specify what
+            service requests that will continue to be processed during
+            draining.
+
+        .. code:: python
+            def exempt_sample(service):
+                if service == 'hello':
```
blampe: return service == 'hello'

blampe: Why do we need this exemption right now, though?

tchannel/tchannel.py
junchaowu: it calls   self._dep_tchannel.advertise_loop.stop() to unadvertise

tchannel/tchannel.py
junchaowu: because we need to be able to relaunch after stop. I am thinking about our test for global tchannel instance.

tchannel/tornado/connection.py
```diff
@@ -122,6 +141,17 @@ def __init__(self, connection, tchannel=None):
 
         self.tchannel = tchannel
 
+        # Map from message ID to inflight requests on the server side.
+        self.incoming_requests = {}
+        # Map from message ID to inflight responses on the client side.
```
blampe: An inbound request on the *client* side would still live in the `inflight_requests` map.

tchannel/tchannel.py
junchaowu: this is matching node implement. Josh told me about it.

tchannel/tornado/connection.py
```diff
@@ -205,50 +235,64 @@ def _loop(self):
...
-                continue
+        if (message.message_type in self.CALL_RES_TYPES and
+                message.flags == FlagsType.fragment):
+            # still streaming, keep it for record
+            self.incoming_responses[message.id] = response
 
-            log.warn('Unconsumed message %s', message)
+    def _handle_response(self, message):
+        future = self._outstanding.pop(message.id)
+        if message.message_type == Types.ERROR:
```
blampe: This error block is copy-pasted.

blampe: Actually most of this function is duplicated logic.

tchannel/tornado/connection.py
junchaowu: An inbound request on the client side, at that time client is regarded as 'server' so it will be in the incoming_requests.

tchannel/tornado/connection.py
```diff
@@ -290,12 +334,7 @@ def write(self, message):
 
         message.id = message.id or self.next_message_id()
 
-        if message.message_type in self.CALL_REQ_TYPES:
-            message_factory = self.request_message_factory
-        else:
-            message_factory = self.response_message_factory
-
-        fragments = message_factory.fragment(message)
+        fragments = self.message_factory.fragment(message)
```
blampe: Nice!

tchannel/tornado/connection.py
```diff
@@ -452,12 +491,26 @@ def serve(self, handler):
         """
         assert handler, "handler is required"
         assert self._loop_running, "Finish the handshake first"
-
         while not self.closed:
             message = yield self.await()
-
             try:
-                handler(message, self)
+                if self._drain:
```
blampe: Way too much branching/duplicated code here.

tchannel/tornado/connection.py
junchaowu: That are not the same. One is working on future obj and the other one is working on response obj. 
If I combine them together, there will be lots of if else.

tchannel/tornado/connection.py
```diff
@@ -489,6 +542,28 @@ def ping(self):
     def pong(self):
         return self._write(messages.PingResponseMessage())
 
+    def drain(self, reason=None, exempt=None):
+        self._drain = _Drain(reason, exempt)
+
+        if len(self.incoming_requests) == 0:
+            self._drain.future.set_result(None)
```
blampe: Call a method on the object that returns a future.

tchannel/tornado/connection.py
```diff
@@ -575,7 +650,7 @@ def stream_request(self, request):
         """send the given request and response is not required"""
         request.close_argstreams()
 
-        stream_future = self._stream(request, self.request_message_factory)
+        stream_future = self._stream(request, self.message_factory)
```
blampe: This is good cleanup.

tchannel/tornado/connection.py
blampe: Right now you're doing high-level logic inside low-level connection logic. The dispatcher is the place where we should be denying inbound requests. You can do this while also keeping the registered handlers.

- Put dispatcher in a state where it starts returning declines
- Call close on all of the peers we want to kill
- Connection makes a note of its outstanding requests/responses and returns a future that resolves when all of those are done
- A timeout callback actually kills the connection after some period of time

tests/test_drain.py
```diff
@@ -0,0 +1,224 @@
...
+    @tornado.gen.coroutine
+    def endpoint(request):
+        raise tornado.gen.Return("hello")
+
+    server.raw.register("endpoint")(endpoint)
+    server.listen()
+    server.stop()
+
+    client = TChannel("client")
+    with pytest.raises(NetworkError):
```
blampe: This would pass if we never started the server, need to assert something stronger.

tests/test_drain.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+@pytest.mark.gen_test
+def test_drain_no_new_connection():
+    server = TChannel("server")
+    server.listen()
+    server.drain()
+
+    client = TChannel("client")
+    with pytest.raises(StreamClosedError):
```
blampe: I don't think this is the error we want to surface to the client. Probably want a network error?

junchaowu: drain() and stop() are different. if tchannel.stop , it will be network error.

tchannel/tchannel.py
blampe: Unadvertise is a call to hyperbahn. Stopping out unad loop means hyperbahn still thinks we're available for the next 5 minutes.

tests/test_drain.py
blampe: The client doesn't care whether you called drain or stop. The client only cares if it had an outstanding request killed, which is its own exception type.

As you've written this, attempting to re-use a dead connection can raise *two* possible exception types depending on what the server did. But it doesn't matter what the server did -- a dead connection is a dead connection, you can't use it.

CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Changes by Version
 0.17.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Add ``TChannel.stop()`` method to gracefully shutdown TChannel server.
```
blampe: This needs to be merged into 0.18.0.

tchannel/tchannel.py
blampe: Still don't understand why this is happening. `relaunch after stop` should be indistinguishable from a normal `listen()`.

tchannel/tchannel.py
```diff
@@ -319,3 +321,36 @@ def advertise(self, routers=None, name=None, timeout=None,
...
+        self._dep_tchannel.peer_group.clear()
+        self._dep_tchannel._server.stop()
+        self._dep_tchannel._server = None
+        self._dep_tchannel._state = State.ready
+        if self.known_peers:
+            for peer_hostport in self.known_peers:
+                self._dep_tchannel.peer_group.add(peer_hostport)
+
+        if self._dep_tchannel.advertise_loop:
+            self._dep_tchannel.advertise_loop.stop()
```
blampe: Is this guaranteed to be running? What if it's not?

tchannel/tornado/connection.py
blampe: Actually how is any of this different from `_outstanding` above?

tchannel/tornado/connection.py
```diff
@@ -205,50 +238,64 @@ def _loop(self):
         #
         # Must be started only after the handshake has been performed.
         self._loop_running = True
+
         while not self.closed:
-            message = yield self._recv()
-            # TODO: There should probably be a try-catch on the yield.
+            try:
+                message = yield self._recv()
+            except tornado.iostream.StreamClosedError:
```
blampe: What does the client see?

tchannel/tornado/connection.py
```diff
@@ -205,50 +238,64 @@ def _loop(self):
...
+                # It is invalid if any CALL_RES_CONTINUE message goes here.
+                self._handle_response(message)
+            elif message.id in self.incoming_responses:
+                # This should just handle CALL_RES_CONTINUE or ERROR message.
+                # It is invalid if any CALL_RES message goes here.
+                self._handle_continuing_response(message)
+            else:
+                log.warn('Unconsumed message %s', message)
+
+    def _handle_continuing_response(self, message):
```
blampe: Why are we still doing all this low-level stuff? Just build the message normally.

tchannel/tchannel.py
junchaowu: If this is not running, _dep_tchannel.advertise_loop will be None.

PR #182, opened 2015-09-18
vcr shouldn't match on hostport

### LINKED ISSUE #175: vcr shouldn't match on hostport
When testing against a service running ephemeral ports it doesn't make sense to care about the hostport.

blampe:
resolves #175

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.17.2 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


blampe:
@breerly @junchaowu

abhinav:
:ok: 

Might be worth removing the hostPort from the VCR IDL to keep it from being recorded at all.

junchaowu:
lgtm

PR #184, opened 2015-09-21
add thrift support to tcurl

### LINKED ISSUE #6: Add Thrift support to tcurl.py.
Right now python services need to awkwardly install node packages to be able to curl their service. Adding thrift support to `tcurl.py` would be easy and would enable this out of the box.

blampe:
Shooting to landing this in the 0.18 branch. Resolves #6.

- Thrift support
- 100% test coverage
- Args more closely match tcurl.js
- Errors only surface as tracebacks with `-v`

@breerly @junchaowu 

```
$ tcurl.py -h
usage: tcurl.py [-h] --service SERVICE [--host HOST] [--endpoint ENDPOINT]
                [--headers HEADERS] [--body BODY] [--timeout TIMEOUT]
                [--health] [-v] [--thrift THRIFT] [--raw]

tcurl: curl for tchannel applications

examples:

    Health check the "larry" service:

      tcurl.py --health --service larry

    Send a Thrift request to "larry":

      tcurl.py --thrift larry.thrift --service larry --endpoint Larry::nyuck \
      --body '{"nyuck": "nyuck"}'

optional arguments:
  -h, --help            show this help message and exit
  --service SERVICE, -s SERVICE
                        Name of destination service for Hyperbahn routing.
                        (default: None)
  --host HOST, -p HOST  Hostname and port to contact. (default:
                        localhost:21300)
  --endpoint ENDPOINT, -1 ENDPOINT
                        Name of destination service for Hyperbahn routing.
                        (default: )
  --headers HEADERS, -2 HEADERS
                        A JSON blob unless --raw was specified, e.g.,
                        --headers '{"foo": "bar"}'. (default: None)
  --body BODY, -3 BODY  A JSON blob unless --raw was specified. This is
                        coerced into a Thrift structure when --thrift is
                        given. (default: None)
  --timeout TIMEOUT     Timeout, in seconds, for the request. (default: 1.0)
  --health              Perform a health check against the given service. This
                        overrides --endpoint. (default: False)
  -v, --verbose         Say more. (default: False)

thrift:
  --thrift THRIFT, -t THRIFT
                        Path to a Thrift IDL file. Incompatible with --raw.
                        (default: None)

raw:
  --raw                 Treats --body as a binary blob instead of JSON.
                        (default: False)
```

### DIFF
--- a/tchannel/tcurl.py
+++ b/tchannel/tcurl.py
@@ -45,14 +45,15 @@ from __future__ import absolute_import
 import argparse
 import logging
 import json
+import os
 import sys
+import traceback
 
 import tornado.ioloop
 import tornado.gen
 
 from . import TChannel
 from . import thrift
-from .health import Meta
 
 log = logging.getLogger('tchannel')
 
@@ -136,6 +137,15 @@ def parse_args(args=None):
         help="Say more.",
     )
 
+    parser.add_argument(
+        "--health",
+        action="store_true",
+        help=(
+            "Perform a health check against the given service. This is "
+            "overridden if --endpoint is provided."
+        ),
+    )
+
     thrift_group = parser.add_argument_group('thrift')
 
     thrift_group.add_argument(
@@ -144,14 +154,6 @@ def parse_args(args=None):
         type=argparse.FileType('r'),
         help=(
             "Path to a Thrift IDL file. Incompatible with --json."
-        ),
-    )
-    thrift_group.add_argument(
-        "--health",
-        action="store_true",
-        help=(
-            "Perform a health check against the given service. This is "
-            "overridden if --endpoint is provided."
         ),
     )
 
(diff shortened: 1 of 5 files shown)


blampe:
@kriskowal fyi

kriskowal:
@blampe A couple things observed from tcurl.js… the JSON flags are dumb. tcurl should just dump plain JSON for as=json and as=thrift, and as=json is default. So, as=raw would be the exception. Use `jq` to fiddle with the output and let tcurl be dumb.

Do you know whether @abhinav left the door open for --no-strict?

junchaowu:
I was thinking about doing this.

tchannel/tcurl.py
```diff
@@ -18,264 +18,259 @@
...
+                hostport=args.host,
+            ),
+            verbose=args.verbose,
+        )
+
+    else:
+
+        result = yield catch_errors(
+            tchannel.raw(
+                service=args.service,
```
junchaowu: We need to set the retry_limit = 0 to stop retry for tcurl.

tchannel/tornado/peer.py
```diff
@@ -519,7 +519,9 @@ def send(
         # peer, we throw exceptions from retry not NoAvailablePeerError.
         peer = self._choose()
         if not peer:
-            raise NoAvailablePeerError("Can't find available peer.")
+            raise NoAvailablePeerError(
+                "Can't find available peer for " + self.service
```
junchaowu: Not sure if this is necessary. Peer are the one we connected to hyperbahn instance. It usually won't be service specific.

junchaowu:
Great work. Just one more point, I am thinking to apply some fuzzy test for tcurl since this will be a tool people use most and with always all kinds of inputs.

blampe:
@breerly @junchaowu can I get another glance before landing.

PR #186, opened 2015-09-21
add some debugging info to no-peer error

blampe:
It doesn't tell you the peer it was trying to find, which is unhelpful.

@junchaowu @breerly

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -519,7 +519,10 @@ class PeerClientOperation(object):
         # peer, we throw exceptions from retry not NoAvailablePeerError.
         peer = self._choose()
         if not peer:
-            raise NoAvailablePeerError("Can't find available peer.")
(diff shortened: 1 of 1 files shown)


tchannel/tornado/peer.py
```diff
@@ -519,7 +519,10 @@ def send(
         # peer, we throw exceptions from retry not NoAvailablePeerError.
         peer = self._choose()
         if not peer:
-            raise NoAvailablePeerError("Can't find available peer.")
+            raise NoAvailablePeerError(
+                "Can't find an available peer for '%s'" % self.service
+            )
```
junchaowu: did you just add this in another pr?

blampe: Yeah breaking it out because I don't want it to wait until 0.18.

blampe:
Gonna land this if no one objects.

junchaowu:
lgtm

PR #190, opened 2015-09-24
Make message_factory stateless

junchaowu:
This is patch that refactor the message_factory class to remove state things out of message_factory class. 

More context for this code change:

The origin message_factory is containing some states information like pending request/response in order to build complete request/response. I want to move those states out of message_factory for couple reasons.

1. make message_factory like util class without needing to remember states.
2. those pending request/response information will be used in many other places like draining connection and peer selection.

The other refactors include removing 'isinstance' methods. Explicitly call build_request or build_response. 
Therefore each connection just needs one message_factory class.

### DIFF
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -33,6 +33,7 @@ from .. import frame
 from .. import glossary
 from .. import messages
 from ..errors import FatalProtocolError
+from ..errors import InvalidChecksumError
 from ..errors import NetworkError
 from ..errors import TChannelError
 from ..event import EventType
@@ -224,8 +225,7 @@ class TornadoConnection(object):
             try:
                 message = yield self._recv()
             except tornado.iostream.StreamClosedError:
-                log.warning("Stream has been closed.")
-                break
+                pass
 
             if message.message_type in self.CALL_REQ_TYPES:
                 self._messages.put(message)
@@ -251,12 +251,14 @@ class TornadoConnection(object):
             )
             return
 
-        self.message_factory.build_inbound_response(message, response)
-
-        if (message.message_type in self.CALL_RES_TYPES and
-                message.flags == FlagsType.fragment):
-            # still streaming, keep it for record
-            self.incoming_responses[message.id] = response
+        try:
+            self.message_factory.build_inbound_response_cont(message, response)
+        except (FatalProtocolError, InvalidChecksumError) as e:
+            response.set_exception(e)
+        else:
+            if message.flags == FlagsType.fragment:
+                # still streaming, keep it for record
+                self.incoming_responses[message.id] = response
 
     def _handle_response(self, message):
         future = self._outstanding.pop(message.id)
@@ -270,13 +272,17 @@ class TornadoConnection(object):
             )
             return
 
-        response = self.message_factory.build_inbound_response(message, None)
-        if future.running():
-            future.set_result(response)
-        if (message.message_type in self.CALL_RES_TYPES and
-                message.flags == FlagsType.fragment):
-            # still streaming, keep it for record
-            self.incoming_responses[message.id] = response
+        try:
+            response = self.message_factory.build_inbound_response(message)
+        except (FatalProtocolError, InvalidChecksumError) as e:
+            if future.running():
+                future.set_exception(e)
+        else:
+            if future.running():
+                future.set_result(response)
+            if message.flags == FlagsType.fragment:
+                # still streaming, keep it for record
+                self.incoming_responses[message.id] = response
 
     # Basically, the only difference between send and write is that send
     # sets up a Future to get the response. That's ideal for peers making
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -43,7 +43,13 @@ from ..messages.common import FlagsType
 from ..messages.common import StreamState
 from ..messages.error import ErrorMessage
 from ..messages.types import Types
-from .message_factory import MessageFactory
+from .message_factory import (
+    build_inbound_error,
+    build_inbound_response,
+    build_inbound_response_cont,
+    build_raw_message,
+    fragment,
+)
 from .util import chain
 
 try:
@@ -103,10 +109,6 @@ class TornadoConnection(object):
 
         # Tracks message IDs for this connection.
         self._id_sequence = 0
-
-        self.message_factory = MessageFactory(
-            self.remote_host, self.remote_host_port
-        )
 
         # Queue of unprocessed incoming calls.
         self._messages = queues.Queue()
@@ -243,7 +245,7 @@ class TornadoConnection(object):
     def _handle_continuing_response(self, message):
         response = self.incoming_responses.pop(message.id)
         if message.message_type == Types.ERROR:
-            error = self.message_factory.build_inbound_error(message)
+            error = build_inbound_error(message)
             response.set_exception(error)
 
             self.tchannel.event_emitter.fire(
@@ -252,7 +254,7 @@ class TornadoConnection(object):
             return
 
         try:
-            self.message_factory.build_inbound_response_cont(message, response)
+            build_inbound_response_cont(message, response)
         except (FatalProtocolError, InvalidChecksumError) as e:
             response.set_exception(e)
         else:
@@ -263,7 +265,7 @@ class TornadoConnection(object):
     def _handle_response(self, message):
         future = self._outstanding.pop(message.id)
         if message.message_type == Types.ERROR:
-            error = self.message_factory.build_inbound_error(message)
+            error = build_inbound_error(message)
             if future.running():
                 future.set_exception(error)
 
@@ -273,7 +275,7 @@ class TornadoConnection(object):
             return
 
         try:
-            response = self.message_factory.build_inbound_response(message)
+            response = build_inbound_response(message)
         except (FatalProtocolError, InvalidChecksumError) as e:
             if future.running():
                 future.set_exception(e)
@@ -309,22 +311,18 @@ class TornadoConnection(object):
 
         future = tornado.gen.Future()
         self._outstanding[message.id] = future
-        self.write(message)
+        self._write(message)
         return future
 
-    def write(self, message):
+    def write(self, fragments):
         """Writes the given message up the wire.
 
         Does not expect a response back for the message.
 
-        :param message:
-            Message to write.
+        :param fragments:
+            A generator that will return fragment messages
         """
         assert not self.closed
-
-        message.id = message.id or self.next_message_id()
-
-        fragments = self.message_factory.fragment(message)
 
         return chain(fragments, self._write)
 
@@ -544,7 +542,7 @@ class StreamConnection(TornadoConnection):
     """
 
     @tornado.gen.coroutine
-    def _stream(self, context, message_factory):
+    def _stream(self, context):
         """write request/response into frames
 
         Transform request/response into protocol level message objects based on
@@ -572,22 +570,21 @@ class StreamConnection(TornadoConnection):
         :param context: Request or Response object
         """
         args = []
+        context.id = context.id or self.next_message_id()
         try:
             for argstream in context.argstreams:
                 chunk = yield argstream.read()
                 args.append(chunk)
                 chunk = yield argstream.read()
                 while chunk:
-                    message = (message_factory.
-                               build_raw_message(context, args))
-                    yield self.write(message)
+                    message = build_raw_message(context, args)
+                    yield self.write(fragment(message, context))
                     args = [chunk]
                     chunk = yield argstream.read()
 
             # last piece of request/response.
-            message = (message_factory.
-                       build_raw_message(context, args, is_completed=True))
-            yield self.write(message)
+            message = build_raw_message(context, args, is_completed=True)
+            yield self.write(fragment(message, context))
             context.state = StreamState.completed
         # Stop streamming immediately if exception occurs on the handler side
         except TChannelError as e:
@@ -598,7 +595,7 @@ class StreamConnection(TornadoConnection):
     def post_response(self, response):
         try:
             # TODO: before_send_response
-            yield self._stream(response, self.message_factory)
+            yield self._stream(response)
 
             # event: send_response
             self.tchannel.event_emitter.fire(
@@ -612,7 +609,7 @@ class StreamConnection(TornadoConnection):
         """send the given request and response is not required"""
         request.close_argstreams()
 
-        stream_future = self._stream(request, self.message_factory)
+        stream_future = self._stream(request)
 
         tornado.ioloop.IOLoop.current().add_future(
             stream_future,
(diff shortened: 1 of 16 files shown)


tchannel/tchannel.py
```diff
@@ -319,3 +319,6 @@ def advertise(self, routers=None, name=None, timeout=None,
         response = Response(json.loads(body), headers or {})
 
         raise gen.Return(response)
+
+    def stop(self, reason=None):
```
breerly: this is a public api dude - please add a thoughtful docblock

breerly:
Can you explain the problem this is fixing?

tchannel/tornado/tchannel.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, name, hostport=None, process_name=None,
         else:
             self._handler = dispatcher
 
-        self.peers = PeerGroup(self)
+        self.peer_group = PeerGroup(self)
```
abhinav: What value does renaming this serve? It makes the PR noisy for no reason.

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
...
-                )
-
-                context.set_exception(error)
-                return error
-        else:
-            return message
+
+    @classmethod
+    def _find_incompleted_stream(cls, reqres):
+        # find the incompleted stream
```
abhinav: incomplete

tchannel/tornado/connection.py
```diff
@@ -122,6 +119,15 @@ def __init__(self, connection, tchannel=None):
...
         self.tchannel = tchannel
 
+        # Map from message ID to inflight requests on the server side.
+        self.incoming_requests = {}
+        # Map from message ID to inflight responses on the client side.
+        # When client sends a request, it will first put ID and a future for
+        # response in the _outstanding map. After it gets response message,
+        # but not a completed response, it will put ID and response obj in the
+        # incoming_response.
+        self.incoming_responses = {}
```
abhinav: Will these point to request and response objects? Which ones? The ones defined in `tchannel.tornado.{request,response}`?

tchannel/tornado/connection.py
```diff
@@ -205,50 +211,64 @@ def _loop(self):
...
         # Must be started only after the handshake has been performed.
         self._loop_running = True
+
         while not self.closed:
-            message = yield self._recv()
-            # TODO: There should probably be a try-catch on the yield.
+            try:
+                message = yield self._recv()
+            except tornado.iostream.StreamClosedError:
+                log.warning("Stream has been closed.")
```
abhinav: This is a useless warning. Drop it. If the stream is closed, that means that the connection should be terminated. We don't need to treat that as a warning.

tchannel/tornado/connection.py
```diff
@@ -205,50 +211,64 @@ def _loop(self):
...
+            return
+
+        self.message_factory.build_inbound_response(message, response)
+
+        if (message.message_type in self.CALL_RES_TYPES and
+                message.flags == FlagsType.fragment):
+            # still streaming, keep it for record
+            self.incoming_responses[message.id] = response
+
+    def _handle_response(self, message):
```
abhinav: might be worth adding an assertion on message type if calling this with anything other than call req/error is an error on our part

tchannel/tornado/connection.py
```diff
@@ -205,50 +211,64 @@ def _loop(self):
...
+                # It is invalid if any CALL_RES_CONTINUE message goes here.
+                self._handle_response(message)
+            elif message.id in self.incoming_responses:
+                # This should just handle CALL_RES_CONTINUE or ERROR message.
+                # It is invalid if any CALL_RES message goes here.
+                self._handle_continuing_response(message)
+            else:
+                log.warn('Unconsumed message %s', message)
+
+    def _handle_continuing_response(self, message):
```
abhinav: same. maybe add assertion on message type

tchannel/tornado/connection.py
abhinav: either that or a catch-all in the if statement below.

tchannel/tornado/connection.py
```diff
@@ -205,50 +211,64 @@ def _loop(self):
...
+
+            self.tchannel.event_emitter.fire(
+                EventType.after_receive_error, error,
+            )
+            return
+
+        response = self.message_factory.build_inbound_response(message, None)
+        if future.running():
+            future.set_result(response)
+        if (message.message_type in self.CALL_RES_TYPES and
```
abhinav: According to your statement above, `message.message_type` MUST BE `CALL_RES` here. You should check for that.

tchannel/tornado/connection.py
```diff
@@ -205,50 +211,64 @@ def _loop(self):
...
+                future.set_exception(error)
+
+            self.tchannel.event_emitter.fire(
+                EventType.after_receive_error, error,
+            )
+            return
+
+        response = self.message_factory.build_inbound_response(message, None)
+        if future.running():
+            future.set_result(response)
```
abhinav: What if response is None because it's not done streaming?

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
...
-        elif message.message_type == Types.CALL_RES:
-            return self.build_response(message)
+    def build_inbound_response(self, message, response):
+        """buffer all the streaming messages based on the
+        message id. Reconstruct all fragments together.
+
+        :param message:
+            incoming message
+        :param response:
+            incoming response
```
abhinav: Explain these parameters further please

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
...
+                src += 1
+
+            if message.flags != FlagsType.fragment:
+                # get last fragment. mark it as completed
+                assert (len(response.argstreams) ==
+                        CallContinueMessage.max_args_num)
+                response.flags = FlagsType.none
+
+            self.close_argstream(response, dst - 1)
+            return None
```
abhinav: add a catch-all else which raises an exception or something if this should never be called with anything besides `CALL_RES*`

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
...
+            response = self.build_response(message)
+            num = self._find_incompleted_stream(response)
+            self.close_argstream(response, num)
+            return response
+
+        elif message.message_type == Types.CALL_RES_CONTINUE:
+            if response is None:
+                # missing call msg before continue msg
+                raise FatalProtocolError(
+                    "missing call message after receiving continue message")
```
abhinav: "Received call response continuation for unknown message %d" % message.id

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
         )
         return res
 
-    def build_context(self, message):
-        if message.message_type == Types.CALL_REQ:
-            return self.build_request(message)
-        elif message.message_type == Types.CALL_RES:
-            return self.build_response(message)
+    def build_inbound_response(self, message, response):
```
abhinav: So if I'm reading this correctly, the state management here is happening like this:

    message = # received new call_res
    response = factory.build_inbound_response(message, None)

    while fragmentation bit is set:
        message = # received new call_res_continue
        factory.build_inbound_response(message, response)

This is inconsistent. `build_inbound_response`'s return value is only used once. This feels like a bad way of managing this state. It would be more consistent if `build_inbound_response` always returned a response but that's pointless if it's mutating the arg stream anyway.

If you have to plumb state like this, maybe message factory should be managing it after all?

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
...
+                # get last fragment. mark it as completed
+                assert (len(response.argstreams) ==
+                        CallContinueMessage.max_args_num)
+                response.flags = FlagsType.none
+
+            self.close_argstream(response, dst - 1)
+            return None
+
+    @classmethod
+    def build_inbound_error(cls, message):
```
abhinav: Same as above

tchannel/tornado/message_factory.py
```diff
@@ -239,94 +236,120 @@ def build_response(self, message):
...
+        """convert error message to TChannelError type."""
+        return TChannelError.from_code(
+            message.code,
+            description=message.description,
+            tracing=message.tracing
+        )
+
+    def build_inbound_request(self, message, request):
```
abhinav: same as above

tchannel/tornado/tchannel.py
junchaowu: Changing the name is because the ``PeerGroup`` class also has the ``peers`` property. If we want to use it, it will look like ``tchannel.peers.peers`` which doesn't make sense. So I change the name.

tchannel/tornado/connection.py
junchaowu: no matter it's streaming or not, when we receive the callresponse message, we will create the response object with all meta information. Even if its streaming, we still will build the response.

junchaowu:
Update most comments from abg, Still double if we need to change message_factory name.

breerly:
@junchaowu no breaking changes? Are you trying to release this as a patch or in `0.18`/`0.19`?

tchannel/tornado/connection.py
```diff
@@ -283,28 +309,19 @@ def send(self, message):
...
-    def write(self, message):
+    def write(self, fragments):
         """Writes the given message up the wire.
 
         Does not expect a response back for the message.
 
-        :param message:
-            Message to write.
+        :param fragments:
+            A generator that will return fragment messages
```
abhinav: This is an API change for TornadoConnection's interface. Are you sure `write` isn't used anywhere else in the library?

If you're really going to change this behavior, also update the docstring and pick a better name for this. Isn't `fragments` just an iterable of messages? Call this `write_all` or something then?

tchannel/tornado/connection.py
```diff
@@ -497,6 +512,16 @@ def ping(self):
...
 
+    def add_incoming_request(self, request):
+        self.incoming_requests[request.id] = request
+
+    def get_incoming_request(self, id):
+        return self.incoming_requests.get(id, None)
+
+    def remove_incoming_request(self, id):
+        req = self.incoming_requests.pop(id, None)
+        return req
```
abhinav: Add docs for all three methods that provide more context. What are the arguments? When/why is this method called?

tchannel/tornado/dispatch.py
```diff
@@ -22,6 +22,8 @@
 
 import logging
 from collections import namedtuple
+from tchannel.tornado.message_factory import build_inbound_request_cont, \
+    build_inbound_request
```
abhinav: nit: avoid `\`. Use `()`.

    from tchannel.tornado.message_factory import (
        build_inbound_request_cont,
        build_inbound_request
    )

abhinav: Or use qualified imports.

    from tchannel.tornado import message_factory

    message_factory.build_inbound_request(...)

tchannel/tornado/dispatch.py
```diff
@@ -71,13 +73,12 @@ def __init__(self, _handler_returns_response=False):
 
     _HANDLER_NAMES = {
         Types.CALL_REQ: 'pre_call',
-        Types.CALL_REQ_CONTINUE: 'pre_call'
+        Types.CALL_REQ_CONTINUE: 'pre_call_cont'
     }
```
abhinav: I don't really like this `getattr` based handling scheme we have going on there. Let's just map message types to methods directly. At the bottom of the class, after all the defs, do:

    _HANDLERS = {
        Types.CALL_REQ: handle_pre_call,
        ...
    }

And in `handle`, do

    self._HANDLERS[message.message_type](self, message, connection)

tchannel/tornado/message_factory.py
```diff
@@ -64,359 +63,377 @@ def build_raw_error_message(protocol_exception):
...
+def build_raw_request_message(request, args, is_completed=False):
+    """build protocol level message based on request and args.
+
+    request object contains meta information about outgoing request.
+    args are the currently chunk data from argstreams
+    is_completed tells the flags of the message
+
+    :param request: Request
+    :param args: array of arg streams
+    :param is_completed: message flags
```
abhinav: `is_completed` is not a message flag. Fix docstring.

tchannel/tornado/tchannel.py
```diff
@@ -396,7 +398,7 @@ def hello_handler(request, response):
 
 
 class TChannelServer(tornado.tcpserver.TCPServer):
-    __slots__ = ('tchannel',)
+    __slots__ = ('tchannel', 'draining')
```
abhinav: draining? you're not using that?

breerly:
Let's talk in person as a team - I'm not ok with merging this before then.

blampe:
Let's table this until we're confident with the state of timeouts.

When we come back to this there are some questions to answer:

 * Why is this necessary? What does this help us achieve? (Is this helping to remove the old API?)
 * What have we done to confirm this is safe?
    * What does coverage look like before/after?
    * What manual steps have been taking to verify correctness?
 * Is there any way to break this into smaller changes?

junchaowu:
As I said so many times, I will explain the reason for this PR for the last time.

Considering TChannel for client/server and request/response, totally it has 4 path. 
1.incoming request path
2.incoming response path
3.outgoing request path
4.outgoing response path

Our old message_factory class uses the following states to manage the workflow of these four path.  And we need two message_factory instances in order to represent request and response types.
```
self.message_buffer = {}	
self.remote_host = remote_host
self.remote_host_port = remote_host_port				
self.in_checksum = {}		
self.out_checksum = {}
```

There are several downside of these.
1. From functionality perspective, the message_factory behavior more like util method class. It provides ways to build request/response or build raw messages. They are more like functions. One place does one thing. 
2. Saving checksum inside the message_factory class's is dangerous. It is dangerous it is not removed after request / response finished processed.
3. Remote host or host port is just used once when they build incoming requests.
4. The message_buffer to save pending incomplete request/response shouldn't be inside the message_factory class. It should belong to connection class. Literally the connection class is responsible to know what's the pending request/response for this connection. Besides, more places will use this pending requests/responses like draining connections, close connections, and calculate peer connection score for peer selections. Also right now we remove the pending incoming request, when we finish building the request. But the user handler may not exit yet. So it is incorrect to remove the pending request at that time.

The change in this diff will solve those downsides.
1. Make all of method as util functions. Use them to build request, response, messages. 
2. Checksum actually is an ongoing things with request/response. Use each request/response's  checksum field to save the ongoing checksum value for future calculation purpose. So we can get rid of a instance level dict to save all req/resp's checksum.
3. Move host/ host_port as function arguments.
4. Move the state of pending request/response into the connection class. So we can just call connection to tell us what's the pending request/response. A pending request should be removed when code exits from user's handler.

Most of code's calculation part remains the same. That's why it doesn't bring too much trouble for testing.

breerly:
holy hell please use markdown to format that comment!

breerly:
@abhinav @junchaowu @blampe can we close this?

abhinav:
Closing. Re-open and fix if this is still needed.

PR #192, opened 2015-09-24
Add binary annotation support for zipkin trace.

junchaowu:
@breerly @blampe 
cc @rf @prashantv

### DIFF
--- a/tchannel/zipkin/annotation.py
+++ b/tchannel/zipkin/annotation.py
@@ -25,7 +25,6 @@ import math
 import time
 
 from tchannel.zipkin.thrift import constants
-from tchannel.zipkin.thrift.ttypes import AnnotationType
 
 Endpoint = collections.namedtuple(
     'Endpoint',
--- a/tchannel/zipkin/annotation.py
+++ b/tchannel/zipkin/annotation.py
@@ -25,6 +25,7 @@ import math
 import time
 
 from tchannel.zipkin.thrift import constants
+from tchannel.zipkin.thrift.ttypes import AnnotationType
 
 Endpoint = collections.namedtuple(
     'Endpoint',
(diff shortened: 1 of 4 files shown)


rf:
lgtm

breerly:
:ship:

blampe:
`Annotation` docstring needs to be updated for the new expected types.

tchannel/zipkin/formatters.py
```diff
@@ -48,6 +48,16 @@
     import json
 
 
+annotation_types = {
```
blampe: We should kill the magic strings and just pass the ttype directly.

tchannel/zipkin/formatters.py
```diff
@@ -124,12 +129,15 @@ def binary_annotation_formatter(annotation):
...
 
-    return ttypes.BinaryAnnotation(
+    ba = ttypes.BinaryAnnotation(
         key=annotation.name,
-        stringValue=value,
-        annotationType=annotation_type
+        annotationType=annotation_type,
     )
 
+    setattr(ba, annotation.annotation_type+'Value', value)
```
blampe: Huh? We should never be calling setattr like this, what are you trying to do?

junchaowu:
```
struct BinaryAnnotation {
    1: string key
    2: optional string stringValue
    3: optional double doubleValue
    4: optional bool boolValue
    5: optional binary bytesValue
    6: optional i64 intValue
    7: AnnotationType annotationType
}
```

the values in binary annotations are named by the pattern like type+'Value'.  So instead of lots of `if-else`, we can just use this way to set value of the binary annotation.

blampe:
@junchaowu Please let us review new diffs before landing. You didn't update the changelog and this isn't really what I had in mind.

junchaowu:
@blampe I revert that. besides the changelog , what's not in your mind.

tchannel/zipkin/annotation.py
```diff
@@ -51,7 +51,7 @@ def __new__(cls, name, value, annotation_type, endpoint=None):
 
         :param annotation_type:
             The expected type of our ``value``. Expected values are
-            ``'string'`` and ``'bytes'``.
+            ``'string'``, ``'int'``, ``'double'`` and ``'bytes'``.
```
abhinav: long?

PR #194, opened 2015-09-26
better benchmark

blampe:
| client | server | rps 
--------- | -------- | ----
| cpython2 | cpython2 | ~700 
| cpython2 | pypy | ~700 
| pypy | cpython2 | ~1000 
| pypy | pypy | ~3500 
| pypy x 2 | pypy | ~4000 

The benchmark is constrained by the client right now, which is why things look better when the client is pypy.

### DIFF
--- a/examples/benchmark/thrift/client.py
+++ b/examples/benchmark/thrift/client.py
@@ -41,6 +41,8 @@ kv = thrift_request_builder(
 local = threading.local()
 local.requests = 0
 
+data = os.urandom(4096)
+
 
 def report_work():
     print local.requests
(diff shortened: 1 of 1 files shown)


examples/benchmark/thrift/client.py
```diff
@@ -50,26 +52,28 @@ def report_work():
...
+    sys.stderr.write('using concurrency %s\n' % concurrency)
+    sys.stderr.flush()
+
+    for _ in xrange(concurrency):
+        do_work()
+
     ioloop.PeriodicCallback(report_work, 1000).start()
+
     try:
         ioloop.IOLoop.current().run_sync(do_work)
```
abhinav: This doesn't seem right. You schedule a bunch of work by calling `do_work` in a loop and then your `run_sync` calls only one instance.

abhinav: You want to block until all the `do_work`s are done, right? You need to use `tornado.gen.multi_future` on the list of futures to get back a future that finishes when all the futures from the list are done. So, something like,

    futures = [do_work() for _ in xrange(concurrency)]
    future = multi_future(futures)

    io_loop.add_future(future, lambda _: io_loop.stop())
    io_loop.start()

blampe: `do_work` does a `while True`, so you have the `run_sync` running one of these loops along with the others already scheduled.

abhinav: Or you can just create another function that does the work and run_sync that in the main block.

blampe:
Landing this unless there are objections.

abhinav:
:-1:

blampe:
NOTED

PR #196, opened 2015-10-05
use thriftrw with vcr

blampe:
WIP

This depends on https://github.com/uber/thriftrw-python/pull/30 (otherwise I would be doing a ton of casting to binary).

Because of https://github.com/uber/thriftrw-python/issues/33 there are also several hacks where I expect a thrift type but get a primitive instead.

Only 4 failures locally, which I think are due to me changing the instance double. 

@abeinstein can you install this branch locally and see if it resolves your issue?

cc @abhinav @breerly @junchaowu


tchannel/testing/vcr/patch.py
```diff
@@ -82,27 +82,34 @@ def send(self, arg1, arg2, arg3,
...
-        vcr_request = VCRProxy.Request(
+        vcr_request = proxy.Request(
             serviceName=self.service.encode('utf-8'),
             hostPort=self.hostport,
             knownPeers=self.original_tchannel.peers.hosts,
             endpoint=endpoint,
             headers=(yield read_full(arg2)),
             body=(yield read_full(arg3)),
-            argScheme=getattr(VCRProxy.ArgScheme, self.arg_scheme.upper()),
+            argScheme=getattr(proxy.module.ArgScheme, self.arg_scheme.upper()),
```
abhinav: You don't need to do `proxy.module.foo`. You can just do `proxy.foo`. `.module` is not a public API. I'll add a `_` later to make that obvious.

tchannel/testing/vcr/cassette.py
```diff
@@ -39,20 +39,55 @@
...
+    if obj and isinstance(obj[0], (str, dict)):
+        return obj
+    return [primitive(v, v.type_spec.ttype_code) for v in obj]
+
+
+def primitive(value, ttype_code):
+    return ({
+        TType.LIST: lambda: primitive_list(value),
+        TType.STRUCT: lambda: primitive_struct(value),
+    }.get(ttype_code) or (lambda: value))()
```
abhinav: Maps and sets?

tchannel/testing/vcr/cassette.py
```diff
@@ -39,20 +39,55 @@
...
+            data[field.name] = primitive(attr, field.ttype_code)
+
+    return data
+
+
+def primitive_list(obj):
+    # I think this is because our objects aren't hydrating with the correct
+    # subtypes.
+    if obj and isinstance(obj[0], (str, dict)):
+        return obj
```
abhinav: Why short circuit on dict? What if dict key/elements are custom types?

tchannel/testing/vcr/cassette.py
```diff
@@ -39,20 +39,55 @@
...
     def to_primitive(self):
+        # import ipdb; ipdb.set_trace()
+        # from thriftrw.protocol import BinaryProtocol
+        # p = BinaryProtocol()
+
+        if hasattr(self.request, 'to_primitive'):
+            return {
+                'request': self.request.to_primitive(),
+                'response': self.response.to_primitive(),
+            }
```
abhinav: I assume this is to work with the old code? You probably want to delete this now.

tchannel/testing/vcr/cassette.py
```diff
@@ -39,20 +39,55 @@
...
+            'request': primitive(self.request, TType.STRUCT),
+            'response': primitive(self.response, TType.STRUCT),
         }
 
     @classmethod
     def to_native(cls, data):
         return cls(
-            request=Request.to_native(data['request']),
-            response=Response.to_native(data['response']),
+            request=proxy.Request(**data['request']),
```
abhinav: this won't handle struct fields, will it?

tchannel/testing/vcr/config.py
```diff
@@ -57,11 +52,11 @@ def __enter__(self):
...
         # TODO Maybe instead of using this instance of the TChannel client, we
         # should use the one being patched to make the requests?
-        client = VCRProxyClient(
-            tchannel=TChannel('proxy-client'),
-            hostport=server.hostport,
-        )
-        self._exit_stack.enter_context(Patcher(client))
+
+        # Need a better way to set hostport after loading!!
+        VCRProxy._module.hostport = server.hostport
```
abhinav: You can just load twice for now. We cache the generated module.

abhinav:
Do you want to just wait for thriftrw to implement `to/from_dict`?

abhinav:
Or is this urgent enough for a custom implementation in tchannel for now?

blampe:
@breerly @abhinav @junchaowu technically this should be backwards compatible but I'm skeptical, should this wait until 0.18? Anyone feel like validating it against some old cassettes? (hi @abeinstein)

breerly:
Sounds like this should be a patch, but it's potentially risky? Why not try releasing it as a patch, if it doesn't work then we can revert as another patch and put it in 0.18.

setup.py
```diff
@@ -18,7 +18,7 @@
         'crcmod',
         'tornado>=4.0,<5.0',
         'toro>=0.8,<0.9',
-        'thriftrw>=0.3,<0.4',
+        'thriftrw>=0.3.3,<0.4.0',
```
breerly: no reason to specify `0.4.0`, `0.4` is the same thing.

Suggest:

```
thriftrw>=0.3.3,<0.4
```

tchannel/testing/vcr/proxy.thrift
```diff
@@ -12,8 +12,8 @@ enum ArgScheme {
 }
 
 struct TransportHeader {
-    1: binary key
-    2: binary value
+    1: required binary key
+    2: required binary value
```
breerly: nice :)

tchannel/testing/vcr/cassette.py
blampe: fixed

tchannel/testing/vcr/cassette.py
blampe: This was because we're hydrating the request from yaml data (like `Request(**data)`) and so nested fields were still just dicts (instead of struct types).

blampe:
Haven't gotten feedback on whether this resolves the VCR issue so we'll probably be able to wait for thriftrw's to/from primitive stuff.

blampe:
@abhinav @breerly @junchaowu another look pls

tchannel/testing/vcr/patch.py
```diff
@@ -82,27 +82,35 @@ def send(self, arg1, arg2, arg3,
         headers = headers or {}
         headers.setdefault('as', self.arg_scheme)
 
-        vcr_request = VCRProxy.Request(
+        vcr_request = proxy.Request(
             serviceName=self.service.encode('utf-8'),
             hostPort=self.hostport,
-            knownPeers=self.original_tchannel.peers.hosts,
+            knownPeers=[bytes(h) for h in self.original_tchannel.peers.hosts],
```
abhinav: thriftrw should take care of this?

tchannel/testing/vcr/patch.py
```diff
@@ -82,27 +82,35 @@ def send(self, arg1, arg2, arg3,
...
-                VCRProxy.TransportHeader(k, v) for k, v in headers.items()
+                proxy.TransportHeader(bytes(k), bytes(v))
+                for k, v in headers.items()
             ],
         )
 
         # TODO what to do with traceflag, attempt-times, ttl
         # TODO catch protocol errors
 
+        # from tchannel import TChannel
```
abhinav: delete

tchannel/testing/vcr/patch.py
```diff
@@ -82,27 +82,35 @@ def send(self, arg1, arg2, arg3,
...
+        from tchannel import TChannel
+        tchannel = TChannel('proxy-client')
+
+        thrift_request = self.vcr_client.send(vcr_request)
```
abhinav: This is not really a client anymore. It's a little confusing to call it that. How about `vcr_service`?

tchannel/testing/vcr/proxy.thrift
```diff
@@ -29,7 +29,7 @@ struct Request {
      */
     5: optional binary hostPort = ""
     6: optional ArgScheme argScheme = ArgScheme.RAW
-    7: optional list<TransportHeader> transportHeaders = {}
```
abhinav: heh. type safety

tchannel/testing/vcr/patch.py
```diff
@@ -110,12 +118,13 @@ def send(self, arg1, arg2, arg3,
                     e.message
                 )
             )
+
         response = Response(
-            code=vcr_response.code,
+            code=vcr_response.status,
```
abhinav: Should probably be `vcr_response.body.code`

abhinav: (the code we want here is the recorded status code, not the status code of making this call to the VCR service)

tests/testing/vcr/test_server.py
```diff
@@ -97,22 +98,22 @@ def f(endpoint, body, headers=None, service=None):
...
+        proxy.Response(
             code=0, headers='{key: value}', body='response body'
         )
     )
 
     response = yield call('endpoint', 'request body')
-    assert response.code == 0
-    assert response.headers == '{key: value}'
-    assert response.body == 'response body'
+    assert response.status == 0
```
abhinav: `response.body.code`

tchannel/testing/vcr/patch.py
blampe: yahp

tchannel/testing/vcr/config.py
abhinav: Created #220

breerly:
:shipit:

breerly:
Update changelog please

abhinav:
:+1:

PR #197, opened 2015-10-05
add changelog for zipkin binary annotation

junchaowu:


breerly:
@junchaowu is this something users should know about? I thought zipkin was transparent to the user.

PR #198, opened 2015-10-05
Use threadloop >=1<2

breerly:
In order to get to 1.0, we need our child dependencies to commit to stable API's - let's start with threadloop which went 1.0 today.


abhinav:
:+1:

blampe:
:+1:

PR #199, opened 2015-10-05
WIP - 0.18.0

### LINKED ISSUE #15: Remove hard dependency on Thrift.
When we added zipkin integration, we somehow added a hard dependency on the Thrift library. We shouldn't have that.

### LINKED ISSUE #117: Json headers should only accept map[string]=string
And should `ValueError` otherwise, just like with out Thrift implementation.

breerly:
Let's add planned 0.18.0 features here:

## Moving towards `thrift.load`

- [x] examples folder should use `thrift.load`
- [x] documentation should use `thrift.load`
- [x] guide should use `thrift.load`

## Deprecations

- [x] add deprecation warnings to api's that will be removed soon - #227 
- [x] remove internal usage of `thrift_request_builder`
- [x] remove internal usage of `client_for`

## Removing implicit dependency on `thrift` package

Resolves #15.

- [x] refactor zipkin to use `thrift.load` #206
- [x] refactor vcr to use `thrift.load` #196
- [x] Drop top-level imports of `thrift` in the serialization logic so that it can be used without having the thrift library installed if thriftrw is being used.

## In preperation for 1.0

- [x] tighen up version constraints in setup.py (only thriftrw remains <1.0)
- [x] json headers should be `map[string]string` instead of any json-serializable (resolves #117)
- [x] remove `tchannel.testing.data` - #225 
- [x] fix possiblity to get socket already in use with port 0.

## New functionality

- [x] add thrift support to tcurl.py
- [x] create optional singleton facility #217 
- [ ] ~~add event hooks to public api (using new req/res, with timers, and content size)~~ (**punted**)


breerly:
These flaky port conflict tests are driving me nuts ;(

junchaowu:
I have two changes coming: 
one is remove the old tchannel handler (request, response)
the other one is refactor the message factory.

breerly:
@junchaowu can you list specifically which classes you want to remove, re: request/response?

breerly:
@junchaowu re:refactor, I don't feel strongly either way.

junchaowu:
right now we still have the endpoint that takes request and response. I want to remove the old one because it is related that requestContext task. I want to put the new request into the request Context but if the old endpoint still exists, that means their will be two different kind of requests that we expose to the user.

breerly:
@junchaowu sure, submit PR! :)

breerly:
@junchaowu my plan is to release this tomorrow evening - anything that won't make that time should probably go in the 0.19 release

junchaowu:
okay

breerly:
@blampe @junchaowu @abhinav hoping to cut this release in the morning - will make sure you guys are onboard before-hand.

PR #200, opened 2015-10-05
Tighten constraints in preparation for 1.0

breerly:
This puts us in a state where the only `>1` dependency is `thriftrw`, which should go `1.0` in 4-6 weeks or so. 

* remove deps duplicated in setup.py and requirements-test.txt
* remove toro, up tornado dep to >=4.2
* remove pyfarmhash
* add version constraint to crcmod
* annotate deps in setup.py into 3 different categories, to keep things explicit and clear


blampe:
lgtm

abhinav:
:+1:

PR #203, opened 2015-10-06
Fix tcurl statements to work with tchannel name

gjtrowbridge:
Make tchannel name and tcurl example statements consistent.


breerly:
Thanks again.

PR #204, opened 2015-10-06
WIP - remove thrift_request_builder

breerly:
- [ ] Remove `thrift_request_builder`
- [ ] Remove `tchannel.testing.data`
- [ ] Upgrade `tchannel.zipkin` to use `thrift.load`


breerly:
On second thought, after talking w @blampe, I'm going to close this and instead explore keeping `thrift_request_builder` around, instead first removing the reliance on `thrift_request_builder` *inside* of TChannel.

PR #205, opened 2015-10-07
change peer selection strategy to prefer incoming conn over outgoing conn

junchaowu:
@abhinav @blampe @breerly @ShanniLi

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -295,9 +295,9 @@ class Peer(object):
         :return:
             A future containing a connection to this host.
         """
-        # Prefer recently created outgoing connections over everything else.
+        # Prefer incoming connection over outgoing connection.
         conns = (
-            conn for conn in chain(self._out_conns, self._in_conns)
+            conn for conn in chain(self._in_conns, self._out_conns)
             if not conn.closed
         )
         try:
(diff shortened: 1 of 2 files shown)


breerly:
Trying to patch `0.17`? Let's update the CHANGELOG if so.

junchaowu:
this is a bugfix and internally.

abhinav:
Regarding changelog: Invisible to the user usually means renaming internal APIs or changing minor internal behavior that doesn't affect anything outside. This change actually affects the user: It fixes a bug which was causing invalid behavior. Consider adding a changelog entry similar to:

* Fixed a bug which caused incoming connections for certain peers to be ignored in favor of outgoing connections.

Also, this change doesn't really fix the problem. Now, we'll just always use the same incoming connection. How about cycling through all connections for a peer instead?

ShanniLi:
@abhinav It is OK to always use one connection. Actual, there will only be one in-connection from Hyperbahn for client to use.

The idea of this PR is choosing the right peer (i.e., Hyperbahn node) for requests.

abhinav:
This PR doesn't change the peer selection logic at all. It changes how multiple connections for a single peer are treated.

junchaowu:
After discuss with juncaoli, this PR will depend on https://github.com/uber/tchannel-python/pull/190 .

breerly:
@junchaowu why is this blocked by that refactor?

junchaowu:
basically to calculate a health weight of a peer depending on some factors like pending requests/response in that peer, number of connections. The pending requests/responses will be part of that PR

junchaowu:
doesn't apply anymore.

PR #208, opened 2015-10-08
Hide more internal details for thriftrw

abhinav:
These should never be accessed directly.

@blampe @breerly @junchaowu


breerly:
Hard to argue with that :clap:

breerly:
:ship:

blampe:
Worth mentioning in the upgrade guide in case anyone was crazy enough to find this already. Otherwise LGTM

abhinav:
Updated

PR #215, opened 2015-10-09
Add matchers argument to use_cassette

abhinav:
This allows customizing the list of request attributes used matched by VCR to
determine if two requests should be considered equal. 

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Changes by Version
 
 - Added Thrift support to ``tcurl.py`` and re-worked the script's arguments.
 - Changed minimum required version of Tornado to 4.2.
+- Added the ability to customize the attributes ``tchannel.testing.vcr`` uses
+  to match requests for replaying responses.
 - **BREAKING** - headers for JSON handlers are not longer JSON blobs but are
   instead maps of strings to strings. This mirrors behavior for Thrift
   handlers.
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -6,6 +6,8 @@ Changes by Version
 
 - Added Thrift support to ``tcurl.py`` and re-worked the script's arguments.
 - Changed minimum required version of Tornado to 4.2.
+- Added the ability to customize the attributes ``tchannel.testing.vcr`` uses
```
breerly: Try "Specify which request components to match on with `tchannel.testing.vcr` - e.g. 'header', 'body', etc..."

abhinav: Updated

tchannel/testing/vcr/config.py
```diff
@@ -141,8 +146,43 @@ def test_bar():
...
+
+        For example,
+
+        .. code-block:: python
+
+            @vcr.use_cassette('tests/data/foo.yaml', matchers=['body']):
+            def test_foo():
+                # ...
+
+            def no_headers(matchers):
```
blampe: Why not a func that returns true/false for each matcher?

```
def match(name):
     return name not in ('headers', 'foo')
```

abhinav: That wouldn't allow adding things that aren't already in the default list of matchers

jcorbin:
This looks like valid python :+1:

blampe:
LGTM

PR #217, opened 2015-10-09
TChannel.get_instance() returns makes TChannel act like a singleton

breerly:
Applications that want to avoid global state can instantiate a `TChannel` object themselves, passing the instance around to where it's needed. 

In some cases, an applications code has no way to avoid global state, and they end up writing their own singleton facility - in this case, as simple & thread-safe `TChannel.get_instance()` function can save them a lot of time and hassle.

- [x] implementation
- [x] tests
- [x] docs

### DIFF
--- a/tchannel/sync/singleton.py
+++ b/tchannel/sync/singleton.py
@@ -25,4 +25,4 @@ class TChannel(TChannelSingleton):
 
         :returns: tchannel.sync.TChannel
         """
-        return supet(TChannel, cls).get_instance()
+        return super(TChannel, cls).get_instance()
--- a/tchannel/sync/singleton.py
+++ b/tchannel/sync/singleton.py
@@ -1,0 +1,12 @@
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+from tchannel.singleton import TChannel as TChannelSingleton
+
+from .client import TChannel as SyncTChannel
+
+
+class TChannel(TChannelSingleton):
+
+    tchannel_cls = SyncTChannel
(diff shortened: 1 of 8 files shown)


abhinav:
So what we want is for people to be able to call `get_instance` on something anywhere in their code without any arguments and get the `TChannel` object. Under the current design, you'd have to pass the correct constructor args everywhere you call `get_instance`. 

As discussed in person, it may be worth having some sort of `prepare` method that accepts the constructor arguments and then everywhere else in the code, calling `get_instance` gets a propertyl configured TChannel instance.

Also, we shouldn't bundle these utility methods all on the TChannel class.

breerly:
@blampe @abhinav @junchaowu 

I've updated the API to address' abg's comments. Now it acts like so:

```
# async

from tchannel.singleton import TChannel

TChannel.prepare(args, kwargs)

tchannel = TChannel.get_instance()

# sync

from tchannel.sync.singleton import TChannel

TChannel.prepare(args, kwargs)

tchannel = TChannel.get_instance()
```

tchannel/sync/singleton.py
```diff
@@ -0,0 +1,12 @@
...
+)
+
+from tchannel.singleton import TChannel as TChannelSingleton
+
+from .client import TChannel as SyncTChannel
+
+
+class TChannel(TChannelSingleton):
+
+    tchannel_cls = SyncTChannel
```
abhinav: Also assign the threadlocal here anew. We don't want sync and async to share the thread local.

junchaowu:
If this is for a client only user, it may make sense. If this is a server, and it is using clay-hyperbahn, it will just use clay-hyperbahn's tchannel which is a global one.

tchannel/sync/singleton.py
breerly: good catch :)

abhinav:
I'm on-board with this API.

abhinav:
:+1:

blampe:
okay :/

PR #218, opened 2015-10-10
tchannel is compatible with thriftrw 0.4

blampe:
Going to cut a new patch with this.

@breerly @junchaowu @abhinav @jacorreia


abhinav:
:+1:

PR #221, opened 2015-10-12
Fix test for absent return values in thrift scheme

abhinav:
thriftrw now raises TypeError if return value is absent for non-void methods.

@breerly @blampe @junchaowu

### DIFF
--- a/tests/schemes/test_thrift.py
+++ b/tests/schemes/test_thrift.py
@@ -1012,7 +1012,7 @@ def test_call_unexpected_error_should_result_in_unexpected_error(
 @pytest.mark.gen_test
 @pytest.mark.call
 def test_value_expected_but_none_returned_should_error(
-    server, service, ThriftTest, use_thriftrw_server
+    server, service, ThriftTest, use_thriftrw_server, use_thriftrw_client
 ):
 
     # Given this test server:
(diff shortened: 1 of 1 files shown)


junchaowu:
lgtm thx for thefix

tests/schemes/test_thrift.py
```diff
@@ -1031,7 +1031,13 @@ def testString(request):
         # values.
         exc = UnexpectedError
     else:
-        exc = ValueExpectedError
+        # If server is using thrift, it will be able to return an invalid
+        # response. thriftrw will fail with a TypeError on invalid values. For
+        # thrift, we'll check manually and raise ValueExpectedError.
+        if use_thriftrw_client:
+            exc = TypeError
```
blampe: Should this be caught and re-raised as ValueExpectedError for consistency?

abhinav: TypeError can also be raised if required arguments in a struct were absent, or on type mismatch. There's no error specific to "return value was absent" raised by thriftrw yet. I can fix that in the next version along with the unexpected exceptions support. Created uber/thriftrw-python#49.

Meanwhile this fix should suffice because this is a very specific corner case that will happen only if the server is using Apache Thrift, the client is using thriftrw, and the server forgets to return a value (which would normally be caught by their tests/manual checks). And when it does occur, there's no good way for the client to handle it except blow up.

blampe: Sure sounds good!

PR #225, opened 2015-10-12
move tchannel.testing.data to tests.data

breerly:
- Move `tchannel.testing.data` out of public library (will remove when we remove Apache Thrift support)
- Refactored test_curl to not use tchannel.testing.data.


abhinav:
:+1:

blampe:
k

PR #229, opened 2015-10-13
vendor tornado's bind_socket for port re-use

blampe:
I'll re-run tests a few time to see if this fixes the issue.

@abhinav @breerly @junchaowu

### DIFF
--- a/tchannel/net.py
+++ b/tchannel/net.py
@@ -21,8 +21,10 @@
 from __future__ import absolute_import
 
 import fcntl
+import os
 import socket
 import struct
+import sys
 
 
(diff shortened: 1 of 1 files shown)


junchaowu:
what problem do you see with bind_socket?

abhinav:
:+1:

junchaowu:
I doubt it will work. After reading this: http://stackoverflow.com/questions/14388706/socket-options-so-reuseaddr-and-so-reuseport-how-do-they-differ-do-they-mean-t

blampe:
Not seeing in-use errors but still getting timeouts. Giving up on this approach for now.

PR #233, opened 2015-10-14
ttl is always ms, convert to seconds only when needed

breerly:


breerly:
shit ... need to create a patch branch for 0.16 first...

tchannel/tornado/connection.py
```diff
@@ -215,6 +215,8 @@ def _loop(self):
         self._loop_running = True
         while not self.closed:
             message = yield self._recv()
+            print message
```
breerly: remove

junchaowu:
abandon

PR #235, opened 2015-10-14
remove reset condition from the ``clear`` method in peer group

junchaowu:
remove reset condition from the ``clear`` method in peer group
at the same time to make the method non-coroutine.
Add connection.closed() check before call close.


abhinav:
k

blampe:
sure

PR #236, opened 2015-10-14
fix timeout unit mismatch issue

junchaowu:


tchannel/tchannel.py
```diff
@@ -268,7 +268,7 @@ def advertise(self, routers, name=None, timeout=None):
             other services will use to make contact with you.
 
         :param timeout:
-            The timeout (in seconds) for the initial advertise attempt.
+            The timeout (in sec) for the initial advertise attempt.
```
abhinav: so this one is unchanged?

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes by Version
 ==================
 
+0.16.7 (unreleased)
+-------------------
+
+- Fixed a bug that the time unit for ttl is unclear. The unit for ttl is set
```
blampe: Not just unclear, timeouts were broken.

tchannel/tchannel.py
blampe: Keep as seconds.

junchaowu: this one is advertise. Advertise timeout by default is 30s.

blampe: Can you confirm it's passing it as 30000?

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
...
+  to millisecond.
```
abhinav: > Fixed a time unit conversion bug that was treating timeouts incorrectly.

PR #237, opened 2015-10-14
fix timeout unit mismatch issue

junchaowu:


CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes by Version
 ==================
 
+0.16.7 (unreleased)
+-------------------
+
+- Fixed a bug that the time unit for ttl is unclear. The unit for ttl is set
+  to millisecond.
```
abhinav: Maybe try

> Fixed a bug that was converting units for timeouts incorrectly.

tchannel/tchannel.py
```diff
@@ -268,7 +268,7 @@ def advertise(self, routers, name=None, timeout=None):
             other services will use to make contact with you.
 
         :param timeout:
-            The timeout (in seconds) for the initial advertise attempt.
+            The timeout (in sec) for the initial advertise attempt.
```
abhinav: why change

tchannel/tornado/message_factory.py
```diff
@@ -97,7 +97,7 @@ def build_raw_request_message(self, request, args, is_completed=False):
         if request.state == StreamState.init:
             message = CallRequestMessage(
                 flags=request.flags,
-                ttl=request.ttl * 1000,
+                ttl=request.ttl,
```
abhinav: can we please use units in the variable names for clarity in the internal APIs?

`ttl_ms` and `ttl_secs` is a lot clearer than `ttl`

junchaowu: right now all of them are ttl_ms. so no need to have these two.

tchannel/tchannel.py
junchaowu: this timeout is not for TTL. it is the advertise timeout. advertise timeout is the advertise behavior not the request.

abhinav:
You need at least one test that uses the top-level API with some timeout in seconds and then verifies that the final `CallRequest` that is sent over the wire has a ttl with the correct timeout in milliseconds. This will prevent this issue from regressing in the future.

tchannel/tornado/message_factory.py
abhinav: No, you misunderstand me. The point is to *always* use unit names in time variables regardless of how many you're mixing together. That way it's clear what kind of value that variable contains/expects. `ttl_ms` and `ttl_secs` were examples. In this case you'd probably have `ttl_ms` everywhere.

tchannel/tchannel.py
abhinav: yeah but why did you change "seconds" to "sec"?

blampe: +1 keep as sec. As mentioned in the other PR also confirm that advertise correctly passes 30000 timeout.

tests/test_tchannel.py
```diff
@@ -296,3 +296,25 @@ def test_listen_duplicate_ports():
...
+    ) as mock_handle:
+        with pytest.raises(TimeoutError):
+            yield tchannel.call(
+                scheme=schemes.RAW,
+                service='server',
+                arg1='endpoint',
+                hostport=server.hostport,
+                timeout=9,
+            )
+        assert mock_handle.call_args[0][1].ttl == 9
```
abhinav: No, you need to check the `ttl` set on the `CallRequestMessage`. That needs to be in milliseconds. Checking on the `Request` object isn't good enough because we could be consistent there while still leaving the `CallRequestMessage` one incorrect -- and that's the one Hyperbahn cares about.

tchannel/tornado/message_factory.py
abhinav: ^Can you address this in a future change?

abhinav:
Okay. @blampe, @breerly?

tchannel/tornado/message_factory.py
junchaowu: yes

breerly:
I think this looks right - please test it by hand then release a patch for both `0.16` and `0.17` releases.

PR #240, opened 2015-10-15
Retry network errors while connecting

### LINKED ISSUE #239: Client should retry all network errors
```
peer = self._choose()
connection = yield peer.connect()
```
in peer.py

This peer select process should be a recursive one until finally get a peer and build connection.

abhinav:
Resolves #239.

@blampe @breerly @junchaowu


blampe:
:+1:

PR #243, opened 2015-10-15
Timeout in seconds, not milliseconds.

breerly:
- update docs to show timeouts as seconds
- create integration test verifying timeout
- update default timeout to be 30 seconds, not 16 minutes

### DIFF
--- a/tchannel/sync/client.py
+++ b/tchannel/sync/client.py
@@ -51,7 +51,7 @@ class TChannel(AsyncTChannel):
         # thrift_service is the result of a call to ``thrift_request_builder``
         future = tchannel.thrift(
             thrift_service.getItem('foo'),
-            timeout=1000,  #  1 second
+            timeout=2,  #  2 seconds
         )
 
         result = future.result()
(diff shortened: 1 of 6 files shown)


abhinav:
:+1:

kriskowal:
Does this preclude ms resolution timeouts?

junchaowu:
change the default value

abhinav:
@junchaowu We don't want to change behavior in 0.17 yet. This just fixes the docs to reflect what the current behavior is.

abhinav:
:+1:

tchannel/sync/client.py
```diff
@@ -51,7 +51,7 @@ class TChannel(AsyncTChannel):
         # thrift_service is the result of a call to ``thrift_request_builder``
         future = tchannel.thrift(
             thrift_service.getItem('foo'),
-            timeout=1000,  #  1 second
+            timeout=2,  #  2 seconds
```
blampe: 0.5

blampe:
Thanks for jumping on this :+1:

tchannel/sync/client.py
breerly: thx

PR #245, opened 2015-10-15
For VCR, use module.hostport to set the hostport

abhinav:
m = thrift.load(..)
    m.hostport = foo

Is the same as

    m = thrift.load(..)
    m.SomeService._module.hostport = foo

Because `m.SomeService._module` is `m`.

@blampe @breerly @junchaowu


abhinav:
Abandoning in favor of #247. Merging to 0.18.

Issue #220, opened 2015-10-12
Changing the `hostport` after thrift.load


abhinav:
Currently there's no way to change the hostport used by the generated module when using `tchannel.thrift.load`. Need to figure out a good way to do this.

abhinav:
Right now, you can change the hostport like so:

    mod = tchannel.thrift.load(...)
    mod.hostport = bar

Thoughts on what a better API for this would look? I don't imagine you'd want to change it on a per-request basis.

We could do something like the following if we decide we're okay with that level of complexity.

    new_mod = mod.with_hostport(bar)

Thoughts? @blampe @breerly @junchaowu

abhinav:
@breerly suggests adding `hostport=` to `tchannel.thrift(..)`. If present, it takes precedence over the hostport specified in `thrift.load`.

    m = thrift.load(..., hostport=foo)
    tchannel.thrift(m.MyService.something(..), hostport=bar)

PR #247, opened 2015-10-15
For VCR, use module.hostport to set the hostport

abhinav:


blampe:
Don't want to expose this as a method? This marries us to the hostport variable name doesn't it?

abhinav:
This PR is just to eliminate a redundancy in how things are being done right now. It's not a behavioral change.

Issue #220 for discussion on how we want people to change it.

abhinav:
Abandoning in of #251.

PR #248, opened 2015-10-16
Replace references to thrift_request_builder with thrift.load

abhinav:
I didn't find any `client_for` references in docs that weren't documenting `client_for` itself

@blampe @breerly @junchaowu


breerly:
Nice!

breerly:
:shipit:

PR #250, opened 2015-10-16
add code to trigger after_send_error event

junchaowu:
Somehow the after_send_error event is missing from current master. Add it.
Add tracing to the error too.

### DIFF
--- a/tchannel/errors.py
+++ b/tchannel/errors.py
@@ -20,6 +20,8 @@
 
 from __future__ import absolute_import
 
+from tchannel.zipkin.trace import Trace
+
 
 #: The request timed out.
 TIMEOUT = 0x01
@@ -77,7 +79,7 @@ class TChannelError(Exception):
         tracing=None,
     ):
         super(TChannelError, self).__init__(description)
-        self.tracing = tracing
+        self.tracing = tracing or Trace()
         self.id = id
         self.description = description
 
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -33,15 +33,14 @@ from .. import frame
 from .. import glossary
 from .. import messages
 from ..errors import NetworkError
-from ..errors import FatalProtocolError
 from ..errors import TChannelError
 from ..event import EventType
 from ..io import BytesIO
 from ..messages.common import PROTOCOL_VERSION
 from ..messages.common import FlagsType
 from ..messages.common import StreamState
-from ..messages.error import ErrorMessage
 from ..messages.types import Types
+from .message_factory import build_raw_error_message
 from .message_factory import MessageFactory
 from .util import chain
 
@@ -471,25 +470,21 @@ class TornadoConnection(object):
                 # TODO Send error frame back
                 logging.exception("Failed to process %s", repr(message))
 
-    def send_error(self, code, description, message_id):
+    def send_error(self, error):
         """Convenience method for writing Error frames up the wire.
 
-        :param code:
-            Error code
-        :param description:
-            Error description
-        :param message_id:
-            Message in response to which this error is being sent
+        :param error:
+            TChannel Error.
         """
-        if code not in ErrorMessage.ERROR_CODES.keys():
-            raise FatalProtocolError(code)
 
-        return self._write(
-            ErrorMessage(
-                code=code,
-                description=description,
-                id=message_id,
-            ),
+        error_message = build_raw_error_message(error)
+        write_future = self._write(error_message)
+        tornado.ioloop.IOLoop.current().add_future(
+            write_future,
+            lambda f: self.tchannel.event_emitter.fire(
+                EventType.after_send_error,
+                error,
+            )
         )
 
     def ping(self):
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -474,13 +474,12 @@ class TornadoConnection(object):
         """Convenience method for writing Error frames up the wire.
 
         :param error:
-            TChannel Error.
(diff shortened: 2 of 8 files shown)


tchannel/errors.py
```diff
@@ -77,7 +79,7 @@ def __init__(
         tracing=None,
     ):
         super(TChannelError, self).__init__(description)
-        self.tracing = tracing
+        self.tracing = tracing or Trace()
```
blampe: Huh? Why?

tchannel/tornado/connection.py
```diff
@@ -471,25 +470,21 @@ def serve(self, handler):
                 # TODO Send error frame back
                 logging.exception("Failed to process %s", repr(message))
 
-    def send_error(self, code, description, message_id):
+    def send_error(self, error):
```
blampe: This is a breaking change, shouldn't land in a patch.

tchannel/errors.py
junchaowu: Error frame has trace field

tchannel/tornado/connection.py
junchaowu: this is internal api. User won't use it.

tchannel/tornado/dispatch.py
```diff
@@ -220,20 +222,22 @@ def handle_call(self, request, connection):
...
         except Exception as e:
-            msg = "An unexpected error has occurred from the handler"
-            log.exception(msg)
+            log.exception("Unexpected Error: {0}".format(e.message))
 
-            response.set_exception(TChannelError(e.message))
+            error = UnexpectedError(
+                description='An unexpected error occurred from the handler',
+                id=request.id,
+                tracing=request.tracing,
```
blampe: What's going on with all the tracing stuff? Seems like that's a real bugfix that belongs in its own patch with its own tests.

tchannel/tornado/connection.py
```diff
@@ -471,25 +470,21 @@ def serve(self, handler):
...
-        return self._write(
-            ErrorMessage(
-                code=code,
-                description=description,
-                id=message_id,
-            ),
+
+        error_message = build_raw_error_message(error)
+        write_future = self._write(error_message)
+        tornado.ioloop.IOLoop.current().add_future(
```
blampe: So there's an important difference between `add_future` like this and `add_done_callback`. `add_future` schedules things such that another coroutine can run between our future finishing and an event hook firing. `add_done_callback` will execute the hook immediately after the future finishes.

Since event hooks tend to be concerned about things like timing, logging, etc., I think we actually *do* want to use `add_done_callback` with our event hooks. @abhinav would you agree with that?

tchannel/tornado/connection.py
blampe: You're right, I thought I was looking at an event hook.

tchannel/tornado/connection.py
```diff
@@ -471,25 +470,21 @@ def serve(self, handler):
...
         """Convenience method for writing Error frames up the wire.
 
-        :param code:
-            Error code
-        :param description:
-            Error description
-        :param message_id:
-            Message in response to which this error is being sent
+        :param error:
+            TChannel Error.
```
blampe: Reference the actual type with :py:class\`path.to.TheClass\`.

tchannel/tornado/connection.py
abhinav: Sounds reasonable

junchaowu: If your puts time counting things into their event hook, will it results incorrect value?

tchannel/tornado/dispatch.py
```diff
@@ -149,14 +147,18 @@ def handle_call(self, request, connection):
...
 
         if request.endpoint in self.handlers and requested_as != expected_as:
-            connection.send_error(
-                ErrorCode.bad_request,
-                "Your serialization was '%s' but the server expected '%s'" % (
-                    requested_as,
-                    expected_as,
+            connection.send_error(BadRequestError(
+                description=(
+                    "Your serialization was '%s' but the server expected '%s'"
```
abhinav: I would just do,

> Server expected a "%s" request but got "%s"

So it'll end up like 'Server expected "thrift" request but got "json"'.

tchannel/tornado/dispatch.py
```diff
@@ -220,20 +222,22 @@ def handle_call(self, request, connection):
...
-                e.message,
-                request.id,
-            )
+            e.tracing = request.tracing
+            e.id = request.id
+            connection.send_error(e)
         except Exception as e:
-            msg = "An unexpected error has occurred from the handler"
-            log.exception(msg)
+            log.exception("Unexpected Error: {0}".format(e.message))
```
abhinav: We don't use `str.format` anywhere else in the library. Let's use `%`-based formatting here too.

tchannel/tornado/connection.py
blampe: Right.

`a().add_done_callback(b)` -> `a b c`

`add_future(a(), b) -> a c b` (potentially)

tchannel/tornado/dispatch.py
junchaowu: I checked this link: http://stackoverflow.com/questions/5082452/python-string-formatting-vs-format

% maybe not good.

abhinav: We've had this discussion before. I'm not aware of any concrete reasons why one is better than the other. 

The reason I'm asking for the change here is solely for consistency.

abhinav:
lgtm

PR #260, opened 2015-10-20
remove explicit dep on futures

breerly:
we already get futures implicitly through threadloop

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -14,6 +14,18 @@ psutil
 # Test all the pythons
 tox
 
+#
+# futures is installed implicitly through threadloop,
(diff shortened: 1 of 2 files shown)


abhinav:
:+1: 

We don't ever directly use the `concurrent.futures` module in our code base.

breerly:
@abhinav wow - looks like we found a bug in threadloop - nice!

breerly:
hmmm... the issue actually appears to be in tox. I'm going to create a workaround that still accomplishes the goal of this PR and add a detailed comment.

junchaowu:
lgtm

abhinav:
:+1:

PR #261, opened 2015-10-20
fix comment using second instead of ms

junchaowu:


junchaowu:
compared with wrong branch.

PR #268, opened 2015-10-22
Prevent all tests from using thrift code generation

abhinav:
@blampe @breerly @junchaowu

### DIFF
--- a/tests/integration/thrift/test_thriftrw.py
+++ b/tests/integration/thrift/test_thriftrw.py
@@ -22,16 +22,12 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 import pytest
 
-from ...util import get_thrift_file
-
 from tchannel import TChannel
-from tchannel import thrift
 
 
 @pytest.fixture
-def keyvalue(tmpdir, request):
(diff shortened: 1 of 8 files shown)


blampe:
https://www.youtube.com/watch?v=mwKSYh-kEHU

tests/integration/thrift/test_thriftrw.py
```diff
@@ -22,16 +22,12 @@
...
-
 from tchannel import TChannel
-from tchannel import thrift
 
 
 @pytest.fixture
-def keyvalue(tmpdir, request):
-    thrift_file = get_thrift_file(tmpdir)
-    return thrift.load(request.node.name, str(thrift_file))
+def keyvalue(thrift_module):
```
junchaowu: this fixture is doing nothing right now. Shall we delete it and make other test just use thrift_module

abhinav: `keyvalue` is more readable than `thrift_module` IMO. I figured we get to avoid `s/keyvalue/thrift_module/` and keep things readable.

PR #270, opened 2015-10-24
reduce zipkin failures to warnings

blampe:
@junchaowu @abhinav @breerly

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.18.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


abhinav:
It makes sense to still include the trace even if you lower this to warn/info. We want to know what exactly went wrong when the failure does occur.

breerly:
This is a good change. :+1: 

Also, seems like this should go in the 0.19 branch, since it's not fixing a bug.

blampe:
Sure, will land on 0.19.

blampe:
Landing in #273.

PR #276, opened 2015-10-26
Improve peer selection performance

abhinav:
This PR contains a series of related commits. Feel free to review them all together or look at the commits separately.

Some notes regarding peer selection itself:

- Peer connections were previously managed in two separate deques which were concatenated together whenever all connections were needed. However, it turns out we always need all connections and rarely check the connection groups separately. So, peers now maintain a single connection list. Incoming connections are added on one side and outgoing on the other side. Connections know which direction they're in (incoming or outgoing) so we can still query based on direction when needed.
- For individual peers, we prefer to use incoming connections over outgoing connections (see also #205).
- Finally, for the actual peer selection logic, instead of assigning each peer a random score for *all* requests and then picking one of them, I changed the logic to pick one of the peers which we have a connection to at random (so just one `randint` call). If we don't have any existing connections, just pick a peer at random. This logic is exactly the same as before but with less complication. We can improve upon this logic based on discussion next/later.

Oh and one more thing: To write the benchmark, I had to change the order in which PeerGroup vs Peer is defined. That's in the first commit ('Add benchmark [..]'). So look at  the 'Optimize linear peer selection logic' commit to see what changes were made to PeerGroup itself. Or check https://github.com/uber/tchannel-python/compare/67ed184f8901c2e69aea6d5a61a0c2e876a40fcc...peer_selection for a diff ignoring that commit.

Performance improvements:
image

Changelog incoming. 

### DIFF
--- a/benchmarks/test_peer_selection.py
+++ b/benchmarks/test_peer_selection.py
@@ -18,7 +18,9 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-from __future__ import absolute_import, unicode_literals, print_function
+from __future__ import (
+    absolute_import, unicode_literals, print_function, division
+)
 
 import mock
 import random
--- a/benchmarks/test_peer_selection.py
+++ b/benchmarks/test_peer_selection.py
@@ -1,0 +1,97 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import, unicode_literals, print_function
+
+import mock
+import random
+
+from tchannel.tornado.peer import (
+    Peer as _Peer,
+    PeerGroup as _PeerGroup,
+)
+
+
+NUM_PEERS = 1000
+
+
+class FakeConnection(object):
+
+    def __init__(self, hostport):
+        self.hostport = hostport
+        self.closed = False
+
+    @classmethod
+    def outgoing(cls, hostport, process_name=None, serve_hostport=None,
+                 handler=None, tchannel=None):
+        return FakeConnection(hostport)
+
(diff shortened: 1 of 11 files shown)


tchannel/tornado/peer.py
```diff
@@ -698,3 +482,167 @@ def clean_up_outgoing_request(request, connection, error):
...
+
+        hosts = self._peers.viewkeys() - blacklist
+        if not hosts:
+            return None
+
+        peers = list(self._connected_peers(hosts))
+        if peers:
+            return peers[random.randint(0, len(peers)-1)]
+        else:
+            hosts = list(hosts)
```
junchaowu: can you bench mark `random.sample(viewkeys, 1)`?

tchannel/tornado/connection.py
```diff
@@ -49,6 +51,13 @@
 log = logging.getLogger('tchannel')
 
 
+#: Sentinel object representing that the connection is outgoing.
+OUTGOING = object()
```
junchaowu: why not use bool?

blampe: A bool is confusing and error-prone... is "True" incoming or outgoing? You'll ask yourself that every time you use it.

tchannel/tornado/peer.py
abhinav: Just checked. It ends up being the same.

blampe: It's a micro optimization:
```
>>> import timeit, random
>>> n = 10000
>>> timeit.timeit('l[random.randint(0, len(l) - 1)]', setup='import random; l=[1] * %s' % n, number=n) / n
1.6440153121948243e-06
>>> timeit.timeit('random.sample(l, 1)', setup='import random; l=[1] * %s' % n, number=n) / n
2.0699024200439455e-06
```

junchaowu: can you add this line as well : `hosts = list(hosts)`

tchannel/tornado/connection.py
abhinav: Yeah. Consider,

    foo(True, False, True)

Versus,

    foo(Incoming, LazyReads, Ephemeral)

With named arguments we can add readability to the former case (`foo(incoming=True, strict_reads=False, ephemeral=True)`) but then sentinel values allow for more than two states (true/false). In this case it doesn't matter much since we'll only ever have two directions, except that maybe sentinel values are slightly more informative (`direction=incoming` vs `incoming=True`).

junchaowu: You still can do
```
incoming = true
outcoming = false

foo(direction=incoming)
```

tchannel/tornado/peer.py
abhinav: I swapped both, `peers[..]` and `hosts[..]` with `random.sample`, dropping the `list(hosts)` for `hosts` (since `random.sample` doesn't care if hosts is a list or not). In multiple runs, the median time was approximately the same (with a variance of about 5-8 microseconds). The rest of the computation is too expensive for this to make any difference.

tchannel/tornado/connection.py
abhinav: Yes, that's also valid. Why do you think it's better? It takes away some of the safety of sentinel values.

benchmarks/test_peer_selection.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import, unicode_literals, print_function
```
breerly: looks like you're missing one? better to just always have all 4 no matter what

abhinav: I don't usually add `division` there, but okay.

blampe:
If peer selection is still uniform I'd be comfortable calling this a patch release. Otherwise lgtm.

junchaowu:
lgtm

abhinav:
Abandoning PR in favor of #279.

PR #279, opened 2015-10-28
Improve peer selection performance (patch)

abhinav:
#276 as a patch on 0.18.


blampe:
+1 for patch, can revert if necessary.

PR #280, opened 2015-10-28
reduce advertise errors to warnings

blampe:
@lucasmarshall @abhinav @breerly @junchaowu

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.18.2 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


abhinav:
:ok:

junchaowu:
Can you also change the log.exceptions in the connection.py? like line ~494

blampe:
@junchaowu that's a legit error, though. The handler is broken in that case. Failing to periodically advertise is recoverable -- the next request(s) should succeed. Hence it's not appropriate as an error.

PR #286, opened 2015-11-02
Separate handler for call req and call req continue message.

junchaowu:


junchaowu:
closed . Diff to wrong branch.

PR #287, opened 2015-11-02
Separate handler for call req and call req continue message.

junchaowu:


tchannel/tornado/dispatch.py
```diff
@@ -70,50 +71,53 @@ def __init__(self, _handler_returns_response=False):
...
         self._handler_returns_response = _handler_returns_response
 
-    _HANDLER_NAMES = {
-        Types.CALL_REQ: 'pre_call',
-        Types.CALL_REQ_CONTINUE: 'pre_call'
-    }
+        self._HANDLERS = {
+            Types.CALL_REQ: self.handle_call_req,
+            Types.CALL_REQ_CONTINUE: self.handle_call_req_cont,
+        }
```
abhinav: You can make this a class attribute.

tchannel/tornado/dispatch.py
```diff
@@ -70,50 +71,53 @@ def __init__(self, _handler_returns_response=False):
...
         try:
             req = connection.request_message_factory.build(message)
-            # message_factory will create Request only when it receives
-            # CallRequestMessage. It will return None, if it receives
-            # CallRequestContinueMessage.
-            if req:
-                self.handle_call(req, connection)
+            self.handle_call(req, connection)
+        except TChannelError as e:
+            log.warn('Received a bad call request message.', exc_info=True)
```
abhinav: Probably worth including the message object in this?

    log.warn("... %s", message, exc_info=True)

tchannel/tornado/dispatch.py
```diff
@@ -70,50 +71,53 @@ def __init__(self, _handler_returns_response=False):
...
+            e.id = message.id
+            e.tracing = req.tracing if req else e.tracing
+            connection.send_error(e)
```
abhinav: When will `req` not be non-None? Now that the handlers are separated, this will be called only when req was a CALL_REQ which means that you'll always get a request back

junchaowu: if exception happens when you do checksum or this request.id is duplicated

tchannel/tornado/dispatch.py
junchaowu: No, because the endpoint handler is instance method, these handle_call_req and handle_call_req_cont can't call instance method if make it class method.

abhinav: No. How you'll do this is:

    def foo(self, arg1, arg2):
        # ...

    HANDLERS = {'x': foo}

Then to call it,

    HANDLERS['x'](self, arg1, arg2)

tchannel/tornado/dispatch.py
```diff
@@ -70,52 +71,55 @@ def __init__(self, _handler_returns_response=False):
...
-
     def handle(self, message, connection):
         # TODO assert that the handshake was already completed
         assert message, "message must not be None"
 
-        if message.message_type not in self._HANDLER_NAMES:
-            # TODO handle this more gracefully
-            raise NotImplementedError("Unexpected message: %s" % str(message))
+        if message.message_type not in self._HANDLERS:
+            raise FatalProtocolError("Unexpected message: %s" % str(message))
```
abhinav: Are you sure we want to kill the server on unexpected messages? It's probably better to log and ignore the message?

(Yeah, we've been using the NotImplementedError version so far but now that we're changing this, might as well fix it.)

tchannel/tornado/dispatch.py
abhinav: Cool. Deleting comment.

abhinav:
Those tests feel like they're just rewriting the code that you just wrote. They assert things about internal details of the class that the caller shouldn't care about. The behavior of the class isn't actually verified. These tests will just be invalidated if there's any refactoring done to the class' internals.

Please  test the behavior from the caller's point of view without mocking and asserting "this internal method was called"? The caller doesn't care about whether `handle_call_req` was called or whether it even exists. That's a detail internal to the class. That's not part of the contract of the class. Only test and verify the public interface of the class.

Note that you didn't have to change anything else but the `handle` method to do this refactor. So, it's a red flag if your tests depend on things added by the refactor. Your tests here should depend only on `handle`, its arguments, and what it returns.

junchaowu:
Remove old tests since the handler methods have been tested in other tests. Add exception test for better coverage.

tchannel/tornado/dispatch.py
```diff
@@ -70,52 +71,55 @@ def __init__(self, _handler_returns_response=False):
...
-            # message_factory will create Request only when it receives
-            # CallRequestMessage. It will return None, if it receives
-            # CallRequestContinueMessage.
-            if req:
-                self.handle_call(req, connection)
+            self.handle_call(req, connection)
+        except TChannelError as e:
+            log.warn('Received a bad call request message.', exc_info=True)
+            e.id = message.id
+            e.tracing = req.tracing if req else e.tracing
```
abhinav: Nit:

    if req:
        e.tracing = req.tracing

tests/test_dispatch.py
```diff
@@ -37,3 +44,9 @@ def test_dispatch():
 
     endpoint = dispatcher.handlers.get("/hello")[0]
     assert endpoint == dummy_endpoint
+
+
+def test_dispatch_unexpected_message():
+    dispatcher = RequestDispatcher()
+    with pytest.raises(FatalProtocolError):
+        dispatcher.handle(CallResponseMessage(), None)
```
abhinav: Maybe assert that the error message includes what we expect? General convention is:

    with pytest.raises(..) as exc_info:
        ...

    assert 'Received a bad call req cont' in str(exc_info)

abhinav:
Couple nits but :ok:

tchannel/tornado/dispatch.py
```diff
@@ -70,52 +71,55 @@ def __init__(self, _handler_returns_response=False):
...
+            connection.request_message_factory.build(message)
         except TChannelError as e:
-            log.warn('Received a bad request.', exc_info=True)
-            if req:
-                e.tracing = req.tracing
+            log.warn('Received a bad call req cont message.', exc_info=True)
+            e.id = message.id
             connection.send_error(e)
 
+    _HANDLERS = {
```
blampe: Nit: I tend to keep these at the top of the class.

blampe:
cool

tchannel/tornado/dispatch.py
abhinav: Can't put that at the top because it references methods that are being defed. If you put it at the top, the methods haven't been def-ed yet and it'll fail with unknown identifier errors.

blampe: TRUTH

Issue #289, opened 2015-11-04
requests over closed connections should be retried


blampe:
```
def send_request(self, request):
    assert not self.closed
```

If the connection is closed the request should be re-tried on another peer.

blampe:
cc @junchaowu @breerly @abhinav @dnathe4th @alexdiao

junchaowu:
the close could happen at any stage. So maybe just remove the assert and capture the network error to do retry.

alexdiao:
Looks like it's fiexed in 0.19.0?

abhinav:
@alexdiao The 0.19 branch was renamed to 0.21 because we had to make two unplanned minor releases. This change was ported back to 0.20.  It's present in [0.20.1](https://github.com/uber/tchannel-python/blob/master/CHANGES.rst#0201-2015-11-12).

But yeah, we can probably close this issue now. Thanks.

alexdiao:
cool let me see if it works.

Issue #285, opened 2015-11-02
potential delay between recieving a response and firing event hooks


blampe:
I'm skeptical of the events we're seeing reported. Right now response logic looks like this:

```python
response_future = connection.send_request(req)

with timeout(response_future, req.ttl):
    try:
        response = yield response_future
    except TChannelError as error:
        # event: after_receive_error
        self.tchannel.event_emitter.fire(
            EventType.after_receive_error, req, error,
        )
        raise
# event: after_receive_response
self.tchannel.event_emitter.fire(
    EventType.after_receive_response, req, response,
)
```

I'm pretty sure this is wrong because the yield statement isn't guaranteed to resume immediately after the future resolves, so there's potential for delay between receiving a response and firing the `after_receive_response` event for it.

More correct (I think) would be this:

```python
response_future = connection.send_request(req)

def fire_events(f):
    if f.exception():
        self.tchannel.event_emitter.fire(
            EventType.after_receive_error, req, error,
        )
        raise f.exception()
    self.tchannel.event_emitter.fire(
        EventType.after_receive_response, req, response,
    )
    
response_future.add_done_callback(fire_events)

with timeout(response_future, req.ttl):
        yield response_future
```
because `add_done_callback` is preemptive and will execute as soon as the future resolves.

junchaowu:
It's true the sequence of callback is add_done_callback > add_future > yield

PR #291, opened 2015-11-05
Remove assertion for connection close instead throwing StreamClosedError

junchaowu:
this is fix for issue: https://github.com/uber/tchannel-python/issues/289

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -412,7 +412,7 @@ class PeerClientOperation(object):
             try:
                 response = yield response_future
             except StreamClosedError as error:
-                error = NetworkError(
+                network_error = NetworkError(
                     id=req.id,
                     description=error.message,
                     tracing=req.tracing,
@@ -421,7 +421,7 @@ class PeerClientOperation(object):
                 self.tchannel.event_emitter.fire(
                     EventType.after_receive_error, req, error,
                 )
-                raise
+                raise network_error
             except TChannelError as error:
                 # event: after_receive_error
                 self.tchannel.event_emitter.fire(
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -27,17 +27,19 @@ import random
 from collections import deque
 from itertools import takewhile, dropwhile
 from tornado import gen
+from tornado.iostream import StreamClosedError
 
 from ..schemes import DEFAULT as DEFAULT_SCHEME
 from ..retry import (
     DEFAULT as DEFAULT_RETRY, DEFAULT_RETRY_LIMIT
 )
-from tchannel.event import EventType
+
 from ..context import get_current_context
 from ..errors import BadRequestError
 from ..errors import NoAvailablePeerError
 from ..errors import TChannelError
 from ..errors import NetworkError
+from ..event import EventType
 from ..glossary import DEFAULT_TIMEOUT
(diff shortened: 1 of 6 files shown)


junchaowu:
https://github.com/uber/tchannel-python/issues/289

tests/integration/test_client_server.py
```diff
@@ -109,3 +110,24 @@ def test_endpoint_not_found(mock_server):
...
+    tchannel = TChannel(name='test')
+
+    class TestHook(EventHook):
+        def before_send_request(self, request):
+            # close the connection
+            peer = tchannel._dep_tchannel.peers.get(mock_server.hostport)
+            peer.close()
+            peer.connections[0].closed = True
+
+    tchannel.hooks.register(TestHook())
```
abhinav: Why don't you just call close right here before making the call instead of using the hook?

junchaowu: If I call close, then we make a call, we will choose peer and  rebuilt a connection.

abhinav: Sorry, I meant what if you stop the server here?

junchaowu: the way iostream calls the callback method is 
```
    def _run_callback(self, callback, *args):
       ....
        with stack_context.NullContext():
            self._pending_callbacks += 1
            self.io_loop.add_callback(wrapper)
```
from the last line, we see the callback won't be called immediately, there is some delay. That's why even if I stop the server, the call back is triggered later.

abhinav: That's not the problem I'm trying to solve here. I'm trying to solve the hook being used to close the connection. Why not,

    mock_server.tchannel._dep_tchannel.close()

That will tell the server to immediately stop listening for new incoming connections, so when the `call` happens two lines below, it'll be trying to hit a hostport that doesn't exist.

abhinav: Also, just to confirm: Is this test just trying to verify that we raise a `NetworkError ` when the destination doesn't exist? Nothing to do with callbacks, right?

tchannel/tornado/peer.py
```diff
@@ -409,6 +411,17 @@ def _send(self, connection, req):
...
+                error = NetworkError(
+                    id=req.id,
+                    description=error.message,
+                    tracing=req.tracing,
+                )
+                # event: after_receive_error
+                self.tchannel.event_emitter.fire(
+                    EventType.after_receive_error, req, error,
+                )
+                raise
```
abhinav: I'm a little confused here. Are you trying to re-raise the `StreamClosedError` you caught or do you want the new `NetworkError` to be raised?

junchaowu: It is will be raised again since we have a try catch from the caller to tell whether we should retry based on the exception or not. 

This part will be refined in this task: https://github.com/uber/tchannel-python/issues/285

tchannel/tornado/peer.py
```diff
@@ -409,6 +411,17 @@ def _send(self, connection, req):
...
+                network_error = NetworkError(
+                    id=req.id,
+                    description=error.message,
+                    tracing=req.tracing,
+                )
+                # event: after_receive_error
+                self.tchannel.event_emitter.fire(
+                    EventType.after_receive_error, req, error,
+                )
+                raise network_error, 'Stream closed error', error
```
abhinav: what

PR #292, opened 2015-11-06
Add tchannel versions information into init req/res headers

junchaowu:
@blampe @breerly @abhinav 
Add version info into header according to the protocol https://github.com/uber/tchannel/blob/master/docs/protocol.md#headers.


tchannel/__init__.py
```diff
@@ -22,6 +22,9 @@
     absolute_import, division, print_function, unicode_literals
 )
 
+__version__ = '0.19.0'
```
breerly: This should be `__VERSION__`

tchannel/glossary.py
```diff
@@ -18,10 +18,22 @@
...
 # CallRequestMessage uses it as the default TTL value for the message.
 DEFAULT_TIMEOUT = 30  # seconds
+
+TCHANNEL_LANGUAGE = 'python'
+
+# version format x.y.z
+TCHANNEL_LANGUAGE_VERSION = '.'.join(map(str, sys.version_info[0:3]))
+
+# version format x.y.z
+TCHANNEL_VERSION = __version__
```
breerly: Please don't duplicate `__VERSION__` - just use it where you need to instead of also having this constant.

tchannel/tornado/connection.py
```diff
@@ -34,11 +35,16 @@
 
 from .. import errors
 from .. import frame
-from .. import glossary
 from .. import messages
 from ..errors import NetworkError
 from ..errors import TChannelError
 from ..event import EventType
+from ..glossary import (
```
breerly: just import the whole glossary

breerly:
tests?

tchannel/glossary.py
junchaowu: But we have protocol version and tchannel version. If I don't specially set it, it will be confusing.

tchannel/tornado/connection.py
```diff
@@ -464,6 +470,9 @@ def outgoing(cls, hostport, process_name=None, serve_hostport=None,
         yield connection.initiate_handshake(headers={
             'host_port': serve_hostport,
             'process_name': process_name,
+            'tchannel_language': TCHANNEL_LANGUAGE,
+            'tchannel_language_version': TCHANNEL_LANGUAGE_VERSION,
+            'tchannel_version': TCHANNEL_VERSION,
```
blampe: Is this consistent with what the other languages are doing?

junchaowu: https://github.com/uber/tchannel/blob/master/docs/protocol.md#headers

tchannel/glossary.py
```diff
@@ -18,10 +18,22 @@
...
 # Message ID 0xffffffff is reserved
 MAX_MESSAGE_ID = 0xfffffffe
 
 # CallRequestMessage uses it as the default TTL value for the message.
 DEFAULT_TIMEOUT = 30  # seconds
+
+TCHANNEL_LANGUAGE = 'python'
+
+# version format x.y.z
+TCHANNEL_LANGUAGE_VERSION = '.'.join(map(str, sys.version_info[0:3]))
```
breerly: I ran this in ipython and got `2.7.10`. This should be `CPython-2.7.10` instead.

breerly:
```
>>> import platform
>>> platform.python_implementation()
'CPython'
```

breerly: Or we can consider making `TCHANNEL_LANGUAGE` the platform string, although that might affect filtering in dashboards since you wouldn't easily be able to see all the python services.

Let's talk about it as a team tomorrow.

breerly: In fact, we might want to know the framework as well - something like:

```
TCHANNEL_LANGUAGE_VERSION = 'Tornado-CPython-2.7.10'
```

breerly: @jcorbin your opinion here would be nice - Python has alot more variability in terms of it's environment than Go or Node. We need to know if it's all of these things in order to actually understand our customers.

breerly: @abhinav @blampe please review my comments here.

blampe: Great catch. Let's keep interpreter/version separate for query-ability. Can also simplify the Python version string:
```
TCHANNEL_LANGUAGE = platform.python_implementation()
TCHANNEL_LANGUAGE_VERSION = platform.python_version()
```

breerly: just talked with @blampe - let's do this

```
TCHANNEL_LANGUAGE = 'python'
TCHANNEL_LANGUAGE_VERSION = platform.python_implementation() + '-' + platform.python_version()
```

tchannel/glossary.py
```diff
@@ -18,10 +18,25 @@
...
-#
 # Message ID 0xffffffff is reserved
 MAX_MESSAGE_ID = 0xfffffffe
 
 # CallRequestMessage uses it as the default TTL value for the message.
 DEFAULT_TIMEOUT = 30  # seconds
+
+TCHANNEL_LANGUAGE = 'python'
+
+# version format x.y.z
```
breerly: change comment to: `# python environment, eg 'CPython-2.7.10'`

breerly:
:ship: looks good

PR #295, opened 2015-11-07
WIP 0.20

junchaowu:
This release will need to wait 4-6 weeks to go out so that 0.18 can sink-in and and people have plenty of time to move to `tchannel.thrift.load`. 

## Removing Deprecated Functionality

- [x] remove `client_for` - #259 
- [x] remove `thrift_request_builder` - #264 
- [ ] update migration guide to move users to `thrift.load`
- [ ] remove `tchannel.tornado.TChannel.register`

## Remove Apache Thrift Support

- [x] remove `testing.data.generated` - #265 
- [x] update tests to not test apache thrift
- [x] Delete `ThriftRequest` in favor of `ThriftRWRequest` (possibly rename the latter)
- [x] Delete `ThriftSerializer` in favor of `ThriftRWSerializer` (possibly rename the latter)
- [x] Delete Apache Thrift endpoint registration code. 
- [x] lint config should not consider generated code #274
- [x] Add "Removing Apache Thrift Support" to the top of the changelog. - #275

## New Functionality

- [ ] public event api (use new req/res objs, resp timer, content size)
- [ ] mocking test helpers (we already have a recorder for integration tests)
- [ ] customizable request context (#145, #143)
- [ ] `TChannel.register` should just be "raw" endpoints (not `as: raw`) that receive the raw bytes, parse them manually as required, and return raw bytes in the Response. We should be able to refactor `TChannel.{thrift,raw,json}.register` to wrap around this.

## Performance

- [ ] fix event hooks not firing immediately #285
- [x] Make peer selection fast. Currently it's taking a *lot* more time than it should. See:

  Function                                      | Samples   | Percentage
  ----------------------------------------------|-----------|-----------
  tchannel/tornado/peer.py:connections (70ms)   | 1753      | 17.8%
  tchannel/tornado/peer.py:choose      (403ms)  | 1421      | 14.4%
  tchannel/tornado/peer.py:score       (189ms)  | 1074      | 10.9%
  tchannel/tornado/peer.py:connected   (105ms)  | 1003      | 10.2%
  tchannel/tornado/peer.py:get         (71ms)   | 644       | 6.5%

  This is partly fixed by #276.

## In Preperation for 1.0

- [x] remove explicit dep on `futures` - #260 
- [x] cut thriftrw 1.0 and update setup.py to `<=1,>2`
- [ ] remove argstreams
- [x] fix lint problems in setup.py, add setup.py to lint - #271

PR #297, opened 2015-11-09
Preparing for 0.20 release.

abhinav:
Other things from #295 have been punted to 0.21.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,10 +1,10 @@
 Changes by Version
 ==================
 
-0.19.1 (unreleased)
+0.20.0 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Support thriftrw 1.0.
 
 
 0.19.0 (2015-11-06)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Changes by Version
 -------------------
 
 - Support thriftrw 1.0.
+- Drop explicit dependency on the ``futures`` library.
 
 
 0.19.0 (2015-11-06)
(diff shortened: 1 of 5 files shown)


abhinav:
@blampe @breerly @junchaowu

CHANGES.rst
```diff
@@ -1,10 +1,11 @@
 Changes by Version
 ==================
 
-0.19.1 (unreleased)
+0.20.0 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- **BREAKING** - At least thriftrw 1.0 is now required.
```
breerly: Is this really breaking from our customers POV?

setup.py
```diff
@@ -22,14 +22,13 @@
...
         'contextlib2',
-        'futures',
 
         # external deps
         'crcmod>=1,<2',
         'tornado>=4.2,<5',
 
         # tchannel deps
-        'thriftrw>=0.4,<0.6',
+        'thriftrw>=1,<2',
```
breerly: This is really exciting :)

CHANGES.rst
abhinav: Some customers are implicitly picking up the wrong versions right now because they have a `thriftrw > foo` constraint in their requirements.txt, and they're seeing a fairly cryptic attribute error when they try to use it. Similarly, I think that when they upgrade to 0.20, if the requirements.txt is out of date, they might end up with thriftrw < 1.0, and get a similar message. I figure it's better to be safe and classify this as breaking with an explicit migration step to ensure they're on the right version.

breerly:
:ship:

junchaowu:
lgtm

Issue #144, opened 2015-08-28
make testing thrift requests easier


blampe:
I define an endpoint and want to test. For raw/json endpoints this is pretty much data-in-data-out:

```
from tchannel import Request

response = my_registered_tchannel_endpoint(Request("foo"))
assert response.body == ...
```

For thrift endpoints this gets complicated because now we need a ThriftRequest instead. How can we make this easier?

aboutdr:
Yah, totally agree, the way to test ThriftRequest involved endpoints is pretty lenthy right now, I am following up this thread to see how we can make it easier?

blampe:
@breerly @junchaowu

blampe:
Would like something like this:
```
from tchannel.testing import Request

response = my_registered_tchannel_endpoint(Request("foo"))
```

abhinav:
This is partly done by #302 and #337. You can now call handler functions with request objects directly:

    response = my_registered_tchannel_endpoint(Request("foo"))

But "foo" needs to the struct representing the arguments of the method. So, if you had,

    void putValue(1: string key, 2: string value)

You still need the `putValue_args(key, value)` struct passed in as the `Request.body` when calling the function directly. We currently don't expose a way to do that. I was considering something like the following for that:

    @tchannel.thrift.register(KeyValue)
    def putValue(request):
        # ...

    putValue.args("key", "value")   # builds a putValue_args object.

    # This will allow you to do:
    putValue(Request(body=putValue.args("key", "value"))

I'm open to other suggestions.

breerly:
@abhinav +1 that interaction looks really good.

abhinav:
Cool. I'll implement that soon.

CC @blampe, @junchaowu 

If you have any objections about the proposed API (see my last comment), raise them now.

breerly:
@abhinav I want to update the client API (for outbound calls) so that we can pass in the same Request object, instead of how we have it now. I'll snag you in person so we can briefly talk about it.

abhinav:
Cool, but that won't affect how we want the `*_args` struct to be exposed.

abhinav:
Okay, after some discussion, this is the ideal use scenario for this:

```python
kv = thrift.load('keyvalue.thrift')

@tchannel.thrift.register(kv.KeyValue)
def getValue(request):
    key = request.body.key
    if ...:
        raise kv.KeyNotFoundException()
    return kv.Value(value)

request = Request(body=kv.KeyValue.getValue(key="foo"))
[...]

PR #302, opened 2015-11-12
Change tchannel.thrift.register to return the unchanged handler.

### LINKED ISSUE #300: Exception is not being raised after adding exception to 'throws' in the IDL
I am having a strange error where the below test fails with InvalidArgument in the throws section in the IDL below, BUT it passes if I remove it. I have confirmed that the exception is being thrown when I step through the code.

Any ideas why this exception would be swallowed?

Thrift version: 0.9.3
I am loading the thrift spec with thrift.load()

===IDL===
```
service DataAggregationService 
{
  ...
  void tripCreatedOrChanged(
    1: TripCreatedOrChangedRequest request
  ) throws (
    1: ServiceError serviceError,
    2: InvalidArgument invalidArgument
  )
}
```

===Handler===
```
@stats.wrapper("servicename.handlers.tripCreatedOrChanged")
@tchannel.thrift.register(servicename.DataAggregationService)
@thrift_handler_wrapper
@coroutine
def tripCreatedOrChanged(request):
    trip_uuid = validate_uuid(request.body.tripUuid)
    yield DataAggregationService.get_and_log_profile_trip_changes(trip_uuid)
```
NOTE: `validate_uuid` raises an InvalidArgument exception if the uuid is not a valid UUID.

===Test===
```
 @mock.patch.object(DataAggregationService, 'get_and_log_profile_trip_changes')
    @gen_test
    def test_trip_created_or_changed_handles_invalid_uuid(self, mock_get_log):
        """Ensure tripCreatedOrChanged raises an InvalidArgument exception if uuid is invalid."""
        request = MockRequest(tripUuid='invalid-uuid')
        with self.assertRaises(InvalidArgument):
            yield data_aggregation.tripCreatedOrChanged(request)

```

Stacktrace:
```
_______________________________________________ TestDataAggregationHandler.test_user_profile_created_or_changed_handles_invalid_uuid _______________________________________________
env/local/lib/python2.7/site-packages/tornado/testing.py:525: in post_coroutine
    timeout=timeout)
env/local/lib/python2.7/site-packages/tornado/ioloop.py:453: in run_sync
    return future_cell[0].result()
env/local/lib/python2.7/site-packages/tornado/concurrent.py:232: in result
    raise_exc_info(self._exc_info)
env/local/lib/python2.7/site-packages/tornado/gen.py:1017: in run
    yielded = self.gen.send(value)
tests/handlers/test_data_aggregation.py:49: in test_user_profile_created_or_changed_handles_invalid_uuid
    yield data_aggregation.userProfileCreatedOrChanged(request)
E   AssertionError: InvalidArgument not raised
```

abhinav:
This allows calling the handler directly with a Request object in unit tests.

Resolves #300.

Users still have to construct Request objects manually and don't usually have
access to the request/response class for the method so #144 is still open, but
this takes one piece of surprising/annoying behavior away.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,8 +9,10 @@ Changes by Version
   ``tchannel.sync.thrift.client_for`` as planned in 0.18.
 - **BREAKING** - Removed ``tchannel.thrift_request_builder`` as
   planned in 0.18.
-- **BREAKING** - ``tchannel.thrift.register`` returns the wrapped function
-  as-is. This allows calling the function diretly for unit tests.
+- **BREAKING** - Support unit testing endpoints by calling the handler
+  functions directly. This is enabled by changing ``tchannel.thrift.register``
+  to return the registered function unmodified. See Upgrade Guide for more
+  details.
 - Reduced Zipkin submission failures to warnings.
 - Limit the size of arg1 to 16KB.
 - Fix bug which prevented requests from being retried if the candidate
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Changes by Version
   ``tchannel.sync.thrift.client_for`` as planned in 0.18.
 - **BREAKING** - Removed ``tchannel.thrift_request_builder`` as
   planned in 0.18.
+- **BREAKING** - ``tchannel.thrift.register`` returns the wrapped function
+  as-is. This allows calling the function diretly for unit tests.
 - Reduced Zipkin submission failures to warnings.
 - Limit the size of arg1 to 16KB.
 - Fix bug which prevented requests from being retried if the candidate
--- a/tests/schemes/test_thrift.py
+++ b/tests/schemes/test_thrift.py
@@ -1045,7 +1045,7 @@ def test_exception_status_code_is_set(server, ThriftTest, server_ttypes):
 # as if they weren't registered at all
 
 
-def make_request(**kwargs):
+def body(**kwargs):
     """Constructs fake Request objects.
 
     The ``_headers`` kwarg may be used to set the header on the Request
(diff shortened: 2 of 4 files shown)


abhinav:
@breerly @blampe @junchaowu

CHANGES.rst
```diff
@@ -9,6 +9,8 @@ Changes by Version
   ``tchannel.sync.thrift.client_for`` as planned in 0.18.
 - **BREAKING** - Removed ``tchannel.thrift_request_builder`` as
   planned in 0.18.
+- **BREAKING** - ``tchannel.thrift.register`` returns the wrapped function
```
breerly: Suggest renaming to "Enable unit testing endpoints by calling them as functions directly. This is enabled by ``tchannel.thrift.register`` returning..." or something like it.

breerly: AKA lead with the message customers will want to read.

tests/schemes/test_thrift.py
```diff
@@ -1037,3 +1038,83 @@ def testException(request):
...
+
+
+def test_void_with_headers_call_directly(server, ThriftTest):
+
+    @server.thrift.register(ThriftTest)
+    def testVoid(request):
+        assert request.headers == {'req': 'header'}
+        return Response(headers={'resp': 'header'})
+
+    resp = testVoid(make_request(_headers={'req': 'header'}))
```
breerly: Tests are documentation. Any reason you can't just instantiate a Request instead of using mock?

CHANGES.rst
abhinav: done

tests/schemes/test_thrift.py
abhinav: Request can be instantiated for the outer struct but the request class representing the arguments is still inaccessible. We'll need mocks for that until we figure out #144.

abhinav: Done

breerly:
Oh Happy Day. :100:

breerly:
:shipit:

PR #304, opened 2015-11-12
cherry pick removing assertion patch

junchaowu:
https://github.com/uber/tchannel-python/pull/291


abhinav:
:ok: after the tests pass

PR #307, opened 2015-11-16
Add test util to directly test requests to user's handler

junchaowu:
This test util will take user's predefined request to find the endpoint and return response.
@blampe @abhinav

### DIFF
--- a/tchannel/testing/dispatch.py
+++ b/tchannel/testing/dispatch.py
@@ -1,0 +1,43 @@
+from __future__ import absolute_import
+
+from tornado import gen
+
+from ..response import response_from_mixed
+
+
+@gen.coroutine
+def handle(tchannel, request):
+    """Test util function to have tchannel instance to take predefined request
+    object.
+
+    .. code-block:: python
+        tchannel = TChannel('example')
+
+        @tchannel.raw.register
+        @gen.coroutine
+        def testString(request):
+            return Response('resp', headers={'resp': 'header'})
+
+        request = Request(
+            body="body",
+            headers={},
+            transport={},
+            endpoint="endpoint",
+        )
+
+        response = yield handle(tchannel, request)
+
+    :param tchannel:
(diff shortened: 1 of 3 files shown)


tchannel/testing/dispatch.py
```diff
@@ -0,0 +1,43 @@
...
+
+from ..response import response_from_mixed
+
+
+@gen.coroutine
+def handle(tchannel, request):
+    """Test util function to have tchannel instance to take predefined request
+    object.
+
+    .. code-block:: python
```
blampe: The example is good but is missing some info. Like "assuming some tchannel handler foo, you can test it like so" and then go into detail like:

``
from ... import handle

def test_foo():
    ...
```

File: tests/testing/test_dispatch.py [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+    return tchannel
+
+
+@pytest.mark.gen_test
+def test_handle_success(server):
+    request = Request(
+        body="hello",
+        endpoint="hello",
+    )
+    response = yield handle(server, request)
```
blampe: It would be more intuitive to be able to call the method being tested with a request.

junchaowu: If they want to call the method directory, then assume they have the request object, then they can just call it.

blampe: Can you add a test for that use case?

junchaowu: that shouldn't in this diff. It is different thing. I don't know if people want's it or not, because if they call the endpoint method directly, the return value from the method is not a tchannel response object.

blampe: That's the primary use case people want, though -- being able to unit test handlers without needed to actually spin up a server.

junchaowu: then this is either thriftrw's code example or weaver's

PR #308, opened 2015-11-17
fix pytest version conflict

junchaowu:
@abhinav @blampe


requirements-test.txt
```diff
@@ -1,7 +1,7 @@
 # Testing and coverage
 pytest<2.8.0
 pytest-cov
-pytest-timeout
+pytest-timeout<1.0.0
```
abhinav: Just try dropping the upper constraint on pytest instead

PR #312, opened 2015-11-19
Add custom peer heap class

### LINKED ISSUE #51: heapify peer selection
We currently iterate through every peer, which is an `O(k)` operation. We should consider keeping these peer scores in a sorted heap and then picking whatever peer is currently on top.

```python
for host in hosts:
    peer = self.get(host)
    score = peer.state.score()

    if score <= score_threshold:
        continue

    if score > chosen_score:
        chosen_peer = peer
        chosen_score = score
```

junchaowu:
The existing python heapq lib doesn't provide the api to update items in the middle of the heap. That's why I implement our own heap. The implementation of the heap refers to the go [heap](https://github.com/golang/go/blob/master/src/container/heap/heap.go).
It provides the basic operations:
* update
* push
* pop
* peek

@blampe @abhinav 

Resolves #51.

### DIFF
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -40,7 +40,8 @@ class HeapOperation(object):
 def init(h):
     # heapify
     n = h.size()
-    for i in six.moves.range(int(math.floor(n/2)) - 1, -1):
+    for i in six.moves.range(int(math.floor(n/2)) - 1, -1, -1):
+        print i
         down(h, i, n)
 
 
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -38,6 +38,7 @@ class HeapOperation(object):
 
 
 def init(h):
+    """Initialize existing object into the heap."""
     # heapify
     n = h.size()
     for i in six.moves.range(int(math.floor(n/2)) - 1, -1, -1):
@@ -45,11 +46,13 @@ def init(h):
 
 
 def push(h, x):
+    """Push a new value into heap."""
     h.push(x)
     up(h, h.size()-1)
 
 
 def pop(h):
+    """Pop the heap value from the heap."""
     n = h.size() - 1
     h.swap(0, n)
     down(h, 0, n)
@@ -57,6 +60,7 @@ def pop(h):
 
 
 def remove(h, i):
+    """Remove the item at position i of the heap."""
     n = h.size() - 1
     if n != i:
         h.swap(i, n)
@@ -67,6 +71,7 @@ def remove(h, i):
 
 
 def fix(h, i):
+    """Rearrange the heap after the item at position i got updated."""
     down(h, i, h.size())
     up(h, i)
 
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -1,3 +1,23 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
 from __future__ import absolute_import
 from __future__ import division
 
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -6,7 +6,12 @@ import six
 
 
 class HeapOperation(object):
-    def cmp(self, i, j):
+    """HeapOperation defines the interface on how to manipulate the heap.
+    Any extension of heap class should implement all the heap operations in
+    order to have complete heap functions.
+    """
+
+    def less(self, i, j):
         """Compare the items in position i and j of the heap.
 
         :param i: the first item's position of the heap list
@@ -27,28 +32,32 @@ class HeapOperation(object):
         """swap items between position i and j of the heap"""
         raise NotImplementedError()
 
+    def size(self):
+        """Return length of the heap."""
+        raise NotImplementedError()
+
 
 def init(h):
     # heapify
-    n = len(h)
+    n = h.size()
     for i in six.moves.range(int(math.floor(n/2)) - 1, -1):
         down(h, i, n)
 
 
 def push(h, x):
     h.push(x)
-    up(h, len(h)-1)
+    up(h, h.size()-1)
 
 
 def pop(h):
-    n = len(h) - 1
+    n = h.size() - 1
     h.swap(0, n)
     down(h, 0, n)
     return h.pop()
 
 
 def remove(h, i):
-    n = len(h) - 1
+    n = h.size() - 1
     if n != i:
         h.swap(i, n)
         down(h, i, n)
@@ -58,14 +67,14 @@ def remove(h, i):
 
 
 def fix(h, i):
-    down(h, i, len(h))
+    down(h, i, h.size())
     up(h, i)
 
 
 def up(h, child):
     while child > 0:
         parent = int(math.floor((child - 1) / 2))
-        if not h.cmp(child, parent):
+        if not h.less(child, parent):
             break
 
         h.swap(parent, child)
@@ -80,10 +89,10 @@ def down(h, parent, n):
 
         min_child = child1
         child2 = child1 + 1
-        if child2 < n and not h.cmp(child1, child2):
+        if child2 < n and not h.less(child1, child2):
             min_child = child2
 
-        if not h.cmp(min_child, parent):
+        if not h.less(min_child, parent):
             break
 
         h.swap(parent, min_child)
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -41,7 +41,6 @@ def init(h):
     # heapify
     n = h.size()
     for i in six.moves.range(int(math.floor(n/2)) - 1, -1, -1):
-        print i
         down(h, i, n)
 
 
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -11,7 +11,7 @@ class HeapOperation(object):
     order to have complete heap functions.
     """
 
-    def less(self, i, j):
+    def lt(self, i, j):
         """Compare the items in position i and j of the heap.
 
         :param i: the first item's position of the heap list
@@ -74,7 +74,7 @@ def fix(h, i):
 def up(h, child):
     while child > 0:
         parent = int(math.floor((child - 1) / 2))
-        if not h.less(child, parent):
+        if not h.lt(child, parent):
             break
 
         h.swap(parent, child)
@@ -89,10 +89,10 @@ def down(h, parent, n):
 
         min_child = child1
         child2 = child1 + 1
-        if child2 < n and not h.less(child1, child2):
+        if child2 < n and not h.lt(child1, child2):
             min_child = child2
 
-        if not h.less(min_child, parent):
+        if not h.lt(min_child, parent):
             break
 
         h.swap(parent, min_child)
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -1,4 +1,8 @@
 from __future__ import absolute_import
+from __future__ import division
+
+import math
+import six
 
 
 class HeapOperation(object):
@@ -27,7 +31,7 @@ class HeapOperation(object):
 def init(h):
     # heapify
     n = len(h)
-    for i in range(n/2 - 1, -1):
+    for i in six.moves.range(int(math.floor(n/2)) - 1, -1):
         down(h, i, n)
 
 
@@ -58,29 +62,29 @@ def fix(h, i):
     up(h, i)
 
 
-def up(h, j):
-    while j > 0:
-        i = (j - 1) / 2
-        if i == j or not h.cmp(j, i):
+def up(h, child):
+    while child > 0:
+        parent = int(math.floor((child - 1) / 2))
+        if not h.cmp(child, parent):
             break
 
-        h.swap(i, j)
-        j = i
+        h.swap(parent, child)
+        child = parent
 
 
-def down(h, i, n):
+def down(h, parent, n):
     while True:
-        j1 = 2*i + 1
-        if j1 >= n or j1 < 0:
+        child1 = 2 * parent + 1
+        if child1 >= n or child1 < 0:
             break
 
-        j = j1
-        j2 = j1 + 1
-        if j2 < n and not h.cmp(j1, j2):
-            j = j2
+        min_child = child1
+        child2 = child1 + 1
+        if child2 < n and not h.cmp(child1, child2):
+            min_child = child2
 
-        if not h.cmp(j, i):
+        if not h.cmp(min_child, parent):
             break
 
-        h.swap(i, j)
-        i = j
+        h.swap(parent, min_child)
+        parent = min_child
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -1,0 +1,86 @@
+from __future__ import absolute_import
+
+
+class HeapOperation(object):
+    def cmp(self, i, j):
+        """Compare the items in position i and j of the heap.
+
+        :param i: the first item's position of the heap list
+        :param j: the second item's position of the heap list
+        :return true if items[i] < items[j], otherwise false.
+        """
+        raise NotImplementedError()
+
+    def push(self, x):
+        """Push a new item into heap"""
+        raise NotImplementedError()
+
+    def pop(self):
+        """Pop an item from the heap"""
+        raise NotImplementedError()
+
+    def swap(self, i, j):
+        """swap items between position i and j of the heap"""
+        raise NotImplementedError()
+
+
+def init(h):
+    # heapify
+    n = len(h)
+    for i in range(n/2 - 1, -1):
+        down(h, i, n)
+
+
+def push(h, x):
+    h.push(x)
+    up(h, len(h)-1)
+
+
+def pop(h):
+    n = len(h) - 1
+    h.swap(0, n)
+    down(h, 0, n)
+    return h.pop()
+
+
+def remove(h, i):
+    n = len(h) - 1
+    if n != i:
+        h.swap(i, n)
+        down(h, i, n)
+        up(h, i)
+
+    return h.pop()
+
+
+def fix(h, i):
+    down(h, i, len(h))
+    up(h, i)
+
+
+def up(h, j):
+    while j > 0:
+        i = (j - 1) / 2
+        if i == j or not h.cmp(j, i):
+            break
+
+        h.swap(i, j)
+        j = i
+
+
+def down(h, i, n):
+    while True:
+        j1 = 2*i + 1
+        if j1 >= n or j1 < 0:
+            break
+
+        j = j1
+        j2 = j1 + 1
+        if j2 < n and not h.cmp(j1, j2):
+            j = j2
+
+        if not h.cmp(j, i):
+            break
+
+        h.swap(i, j)
+        i = j
--- a/tchannel/peer_heap.py
+++ b/tchannel/peer_heap.py
@@ -1,0 +1,61 @@
+from __future__ import absolute_import
+
+from .container import heap
+from .container.heap import HeapOperation
+
+
+class PeerHeap(HeapOperation):
+    def __init__(self):
+        self.peers = []
+
+    def cmp(self, i, j):
+        return self.peers[i] < self.peers[j]
+
+    def push(self, x):
+        x.index = len(self.peers)
+        self.peers.append(x)
+
+    def pop(self):
+        n = len(self.peers)
+        item = self.peers[n-1]
+        item.index = -1
+        del self.peers[n - 1]
+        return item
+
+    def __len__(self):
+        return len(self.peers)
+
+    def swap(self, i, j):
+        self.peers[i], self.peers[j] = self.peers[j], self.peers[i]
+        self.peers[i].index = i
+        self.peers[j].index = j
+
+    def update_peer(self, peer):
+        """Update the peer's position in the heap after peer's score changed"""
+        heap.fix(self, peer.index)
+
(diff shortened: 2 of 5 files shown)


tchannel/tornado/peer.py
```diff
@@ -96,6 +100,15 @@ def __init__(self, tchannel, hostport):
...
 
+        # score is used to measure the performance of the peer.
+        # It will be used in the peer heap.
+        self.score = sys.maxint
+        # index records the position of the peer in the peer heap
+        self.index = -1
+
+    def __lt__(self, other):
+        return self.score < other.score
+
```
abhinav: As I mentioned yesterday, don't implement `__lt__` on something simply to compare scores conveniently. Unless the concept represented by that class has a semantic meaning of "this peer is less than that peer", they shouldn't be comparable. You're implementing the heap specifically for this use case. You can compare the score with `.score` instead of `x < y`.

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def swap(self, i, j):
+        """swap items between position i and j of the heap"""
+        raise NotImplementedError()
+
+
+def init(h):
+    # heapify
+    n = len(h)
+    for i in range(n/2 - 1, -1):
```
abhinav: Are you sure this works? `range(x, -1)` for `x >= 0` is an empty list.

abhinav: Also, use xrange, not range. Range generates a list in Python 2. 

(Although, preferably you want to use `six.moves.range` if you want to support both, Python 2 and 3.)

abhinav: Also, this code will break in unexpected ways if you're in Python 3. Can you import `division` from `__future__` and then explicitly `math.floor` or `math.ceil` the result of the division?

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import absolute_import
+
+
+class HeapOperation(object):
+    def cmp(self, i, j):
```
abhinav: You need to document what this class does/provides and how this whole module is supposed to be used

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+
+def down(h, i, n):
+    while True:
+        j1 = 2*i + 1
+        if j1 >= n or j1 < 0:
+            break
+
+        j = j1
+        j2 = j1 + 1
+        if j2 < n and not h.cmp(j1, j2):
```
abhinav: Use better variable names please.

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class HeapOperation(object):
+    def cmp(self, i, j):
+        """Compare the items in position i and j of the heap.
+
+        :param i: the first item's position of the heap list
+        :param j: the second item's position of the heap list
+        :return true if items[i] < items[j], otherwise false.
+        """
```
abhinav: cmp implies that you're doing the [`cmp`](https://docs.python.org/2/library/functions.html#cmp) operation. Call this `lt` or actually use `cmp` (in which case `if h.cmp(a, b)` becomes `if h.cmp(a, b) < 0`).

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
...
+from .container import heap
+from .container.heap import HeapOperation
+
+
+class PeerHeap(HeapOperation):
+    def __init__(self):
+        self.peers = []
+
+    def cmp(self, i, j):
+        return self.peers[i] < self.peers[j]
```
abhinav: what's stopping you from comparing the scores here instead of using `__lt__`?

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import absolute_import
+
+from .container import heap
+from .container.heap import HeapOperation
+
+
+class PeerHeap(HeapOperation):
+    def __init__(self):
+        self.peers = []
```
abhinav: Since you'll be adding and removing items from this all the time, consider using [deque](https://docs.python.org/2/library/collections.html#collections.deque)

tests/test_peer_heap.py
```diff
@@ -0,0 +1,97 @@
...
+        peers.append(mock_peer(i))
+
+    random.shuffle(peers)
+    return peers
+
+
+def test_push(peer_heap):
+    peers = mock_peers(n)
+    for peer in peers:
+        peer_heap.push_peer(peer)
```
abhinav: All these tests need to assert that the heap invariant holds after each operation.

abhinav: Also, the tests for the heap implementation itself should probably be separate. Test the heap implementation with a custom HeapOperation. Then test the behavior of the PeerHeap separately. It's okay if there's some overlap in the tests.

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+def init(h):
+    # heapify
+    n = len(h)
+    for i in range(n/2 - 1, -1):
+        down(h, i, n)
+
+
+def push(h, x):
+    h.push(x)
+    up(h, len(h)-1)
```
abhinav: Given that the HeapOperation has custom swap/compare/etc., it should also have a custom length operation so that you don't have to implement `__len__`.

tchannel/peer_heap.py
junchaowu: Oh that's because this is just the simplest version. The real version won't be just compare score only.  Those changes will appear in the future diffs .

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+        :return true if items[i] < items[j], otherwise false.
+        """
+        raise NotImplementedError()
+
+    def push(self, x):
+        """Push a new item into heap"""
+        raise NotImplementedError()
+
+    def pop(self):
+        """Pop an item from the heap"""
```
abhinav: operation naming nit: push suggests you're pushing something to index 0. I would use 'append'.

tchannel/peer_heap.py
junchaowu: There won't be pop action all the time. The whole implementation of heap is depending on list. List has instance access in case we want to update score for some peers directly.

tchannel/tornado/peer.py
junchaowu: explained above.

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+
+def fix(h, i):
+    down(h, i, len(h))
+    up(h, i)
+
+
+def up(h, j):
+    while j > 0:
+        i = (j - 1) / 2
+        if i == j or not h.cmp(j, i):
```
abhinav: Under what case does `i == j` happen?

Also, if `i == j`, `h.cmp(j, i)` will be false anyway.

tchannel/container/heap.py
junchaowu: a custom length operation is the same as implement __len__ right?

tchannel/container/heap.py
junchaowu: good catch. It's unnecessary

tchannel/container/heap.py
```diff
@@ -0,0 +1,86 @@
...
+        h.swap(i, n)
+        down(h, i, n)
+        up(h, i)
+
+    return h.pop()
+
+
+def fix(h, i):
+    down(h, i, len(h))
+    up(h, i)
```
abhinav: Wait, if it turns out that you ended up sifting down, that means that you don't need to sift up, right?

Like if you have something where the heap property is already true, and then you change an item so that it is bigger than its children. `down` will swap it with the smaller of its two children. Now that new value at the index of the changed value doesn't need to be sifted up because we know that the heap property was true before, so the new value is greater than the parent of the value that changed.

Like if you have,

              i-1/2               i            i*2+1   i*2+2
    x = [..., parent_parent, ..., parent, ..., child1, child2, ...]

The heap property is already true so

    parent_parent < parent
    parent < child1
    parent < child2

which implies,

    parent_parent < child1
    parent_parent < child2

If you now change parent to violate the heap property so that it gets sifted down, that is, you change parent to be > child1 or child2, then,

              i-1/2               i            i*2+1   i*2+2
    x = [..., parent_parent, ..., child1, ..., parent, child2, ...]

Since only `parent` changed, `parent_parent < child1` is still true, so you don't need to `up(h, i)` anymore.

tchannel/container/heap.py
junchaowu: if j == 1 the case happens. 
those are two condition in "or" operation

tchannel/peer_heap.py
abhinav: And still, because this is specific to peers, you could put any complex logic/calling a function on peers here. Peers don't need the `__lt__` method.

tchannel/tornado/peer.py
abhinav: I disagree. I still don't see why you can't use `.score` directly.

tchannel/container/heap.py
junchaowu: .... up(h,i) is for the case that you change your parent to be less than parent_parent.  then you need to shift it up.

tchannel/peer_heap.py
abhinav: True. I didn't realize deque isn't good for random access. Apparently it's O(n) for random access.

tchannel/container/heap.py
abhinav: Yes. But only one of the two things happens at a time. I'm saying that if sifting down happened, then sifting up won't happen. If sifting down didn't happen, then sifting up will happen.

tchannel/container/heap.py
junchaowu: I want to make it flexible for minimum heap or maximum heap. If I named it to 'lt' which will imply this is minimum heap.

tchannel/container/heap.py
junchaowu: Yes, one will happen at a time. The other is called but exist immediately

tchannel/container/heap.py
abhinav: I was trying to make two points:

1. Given `j > 0`, it's impossible for `(j - 1) / 2 = j`. Just analytically, if you solve `j = (j - 1) / 2` for `j`, you get `j = -1`. The only time this condition will be hit is when `j = -1` and that's impossible because you have a `j > 0` condition on the while loop.
2. If by some magic `i == j`, then `h.cmp(i, j)` is always false. So, the second part of the `if` condition still holds.

Either way, the `i == j` check isn't required.

tchannel/container/heap.py
abhinav: No, what I'm trying to avoid is magic methods like `__len__`, etc. on things that may not have any definition of that. A peer cannot be less than another peer. Maybe `__len__` on a `PeerHeap` is okay because you want to determine the number of objects inside the heap.

But I would rather not have the implementation of the heap algorithm itself depend on magic methods. `HeapOperation` is the algorithm's interface to the underlying storage. It should not rely on anything implemented on the `HeapOperation` besides what was specified in the interface above. Make a `size`/`length` method a part of the `HeapOperation` interface and use that (`h.size()`/`h.length()`) instead of `len(h)`

tchannel/container/heap.py
abhinav: The `HeapOperation` can always flip its definition of what `lt` or `cmp` means if it wants a min/max heap. This issue is mostly about naming. If you're calling it `cmp` I would expect it to be the Python `cmp` operation: returns value < 0 if left < right, value = 0 if left = right, and value > 0 if left > right.

tchannel/peer_heap.py
junchaowu: python provides this feature. why not use it. I don't see any strong reason I can't use it for peer

tchannel/container/heap.py
abhinav: But it won't exit immediately. It'll still walk up the tree. Sure, it's only `O(logn)` operations in the worst case, but it's obvious that it's unnecessary and it's trivial to avoid.

junchaowu: It will exit. Why do u say it will walk up the tree.

tchannel/peer_heap.py
abhinav: Because there's no semantic meaning to a peer being less than another peer. Availability of a feature doesn't mean it's open to misuse. For things like operator overloading, it's very easy to end up with code that is not obvious by looking at it. That's why you should use operator overloading only when there's an actual meaning of the comparison on the object being compared. If you were comparing a vector, there's a specific meaning to "vector A is less than B", so it's okay to define `__lt__` there.

tchannel/container/heap.py
abhinav: It will walk up the tree without making any changes. It' ll check that `parent(i)` is less than the new `i`, and that `parent(parent(i))` is less than that, and so on until it reaches the root node. All these conditions will be true because of previously explained reasons. These comparisons are very easy to avoid by checking if `down()` sifted the value down, and if so, don't call `up()`.

junchaowu: if it is greater than parent , then it breaks. If it is less then parent, it continue.

abhinav: Ah, okay. Missed the 'break'. Makes sense.

tchannel/container/heap.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def swap(self, i, j):
+        """swap items between position i and j of the heap"""
+        raise NotImplementedError()
+
+
+def init(h):
+    # heapify
+    n = len(h)
+    for i in six.moves.range(int(math.floor(n/2)) - 1, -1):
```
abhinav: Because GitHub hid the comment, just a reminder:

     range(n, -1) == []` for n >= 0

You'll need comprehensive separate tests for the heap implementation.

tchannel/container/heap.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def size(self):
+        """Return length of the heap."""
+        raise NotImplementedError()
+
+
+def init(h):
+    # heapify
+    n = h.size()
+    for i in six.moves.range(int(math.floor(n/2)) - 1, -1):
```
abhinav: FYI this is still broken.

tchannel/container/heap.py
```diff
@@ -0,0 +1,99 @@
...
+import six
+
+
+class HeapOperation(object):
+    """HeapOperation defines the interface on how to manipulate the heap.
+    Any extension of heap class should implement all the heap operations in
+    order to have complete heap functions.
+    """
+
+    def less(self, i, j):
```
abhinav: nit: `less_than`

tests/test_peer_heap.py
```diff
@@ -0,0 +1,110 @@
...
+
+def verify(ph, parent):
+    child1 = 2*parent + 1
+    child2 = 2*parent + 2
+    if child2 < ph.size():
+        assert not ph.less(child1, parent)
+        verify(ph, child1)
+
+    if child2 < ph.size():
+        assert not ph.less(child2, parent)
```
abhinav: nit: 

    assert ph.less(parent, child1)

tchannel/container/heap.py
```diff
@@ -0,0 +1,100 @@
...
+    def size(self):
+        """Return length of the heap."""
+        raise NotImplementedError()
+
+
+def init(h):
+    # heapify
+    n = h.size()
+    for i in six.moves.range(int(math.floor(n/2)) - 1, -1, -1):
+        print i
```
abhinav: ?

junchaowu: oh. I'll delete

tchannel/peer_heap.py
abhinav: Probably add `__slots__` so that we're not paying the cost of a `dict` here.

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def push(self, x):
+        x.index = len(self.peers)
+        self.peers.append(x)
+
+    def pop(self):
+        n = len(self.peers)
+        item = self.peers[n-1]
+        item.index = -1
+        del self.peers[n - 1]
```
abhinav: This can be compressed into:

    item = self.peers.pop()
    item.index = -1
    return item

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import absolute_import
+
+from .container import heap
+from .container.heap import HeapOperation
+
+
+class PeerHeap(HeapOperation):
```
abhinav: Document that this class is a min-heap of peers based on their score.

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
...
+    def pop_peer(self):
+        """Pop the top peer from the heap
+
+        :return
+            return the top peer and remove it from the heap if heap is not
+            empty.Otherwise return None.
+        """
+        if len(self.peers) == 0:
+            return None
+        return heap.pop(self)
```
abhinav: So the public interface of the `PeerHeap` class is: `{push,pop,peek,update}_peer`, right? The other methods are for the `HeapOperation`. You don't need to fix this right now (add a TODO) but this could be two separate classes.  One that's the actual min-heap and the other the public interface for PeerHeap that wraps the min-heap based on the heap.py.

tchannel/container/heap.py
```diff
@@ -0,0 +1,99 @@
+from __future__ import absolute_import
+from __future__ import division
```
abhinav: No license header. Run scripts/install-hooks.sh in the root directory of the project to get the pre-commit hook that adds the license header.

tchannel/container/heap.py
```diff
@@ -0,0 +1,99 @@
...
+    if n != i:
+        h.swap(i, n)
+        down(h, i, n)
+        up(h, i)
+
+    return h.pop()
+
+
+def fix(h, i):
+    down(h, i, h.size())
```
abhinav: Looks like everything except `up` and `down` is the public interface. Can you document the public interface of this module?

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def pop_peer(self):
+        """Pop the top peer from the heap
+
+        :return
+            return the top peer and remove it from the heap if heap is not
+            empty.Otherwise return None.
+        """
+        if len(self.peers) == 0:
+            return None
```
abhinav: Test for truthiness instead

    if not self.peers:
        # ...

tchannel/tornado/peer.py
```diff
@@ -96,6 +100,12 @@ def __init__(self, tchannel, hostport):
         # helps avoid making multiple outgoing connections.
         self._connecting = None
 
+        # score is used to measure the performance of the peer.
+        # It will be used in the peer heap.
+        self.score = sys.maxint
+        # index records the position of the peer in the peer heap
```
abhinav: Why is the default score maxint? If anything it should be 0 since you're not connected to the peer initially.

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+        n = len(self.values)
+        item = self.values[n - 1]
+        del self.values[n - 1]
+        return item
+
+    def swap(self, i, j):
+        self.values[i], self.values[j] = self.values[j], self.values[i]
+
+
+n = 1000
```
abhinav: What's this constant? What does it do? Why not use a fixture?

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+    def lt(self, i, j):
+        return self.values[i] < self.values[j]
+
+    def push(self, x):
+        self.values.append(x)
+
+    def pop(self):
+        n = len(self.values)
+        item = self.values[n - 1]
+        del self.values[n - 1]
```
abhinav: return values.pop()

abhinav:
Can you post the coverage numbers for the new modules?

We should fix coveralls at some point.

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+def int_heap():
+    return IntHeap()
+
+
+def create_values(n):
+    values = []
+    for i in six.moves.range(n):
+        values.append(i)
+
+    random.shuffle(values)
```
abhinav: Let's make this slightly more than "all numbers from 1 to N, each occurring exactly once".

I would create a fixture that represents the list of numbers for that test run. Parameterize it so that you try the tests against different collections of numbers. Try to have at least the following

- What you have right now. Just 1000 numbers with no duplication.
- 100 random numbers between 1 and 1000 (so that there are many duplicates).
- Numbers from opposite ends -- some very small numbers, some very large numbers.

Then in the tests where you're doing `create_values(n)`, just use that fixture instead. The test will be run multiple times with the different lists.

See http://pytest.org/latest/example/parametrize.html#indirect-parametrization-with-multiple-fixtures.

tchannel/peer_heap.py
```diff
@@ -0,0 +1,61 @@
...
+
+class PeerHeap(HeapOperation):
+    def __init__(self):
+        self.peers = []
+
+    def size(self):
+        return len(self.peers)
+
+    def lt(self, i, j):
+        return self.peers[i].score < self.peers[j].score
```
abhinav: This is mostly about naming:

Low score is better? Wasn't it the opposite when we were doing scores before? High score was for more desirable peers. Maybe the attribute should be renamed? "load"? It's some measure of how much load we think that peer is under, so as a load balancer, we pick the one we think is under the least load?

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+    return values
+
+
+def verify(ph, parent):
+    child1 = 2 * parent + 1
```
abhinav: Add a comment or docstring that says,

> Asserts that the heap property still holds.

Tests don't have to be devoid of docs. :)

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def test_pop(int_heap):
+    values = create_values(n)
+    for value in values:
+        heap.push(int_heap, value)
+        verify(int_heap, 0)
+
+    for i in six.moves.range(n):
+        assert i == heap.pop(int_heap)
```
abhinav: check invariant as you pop

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+    values = create_values(n)
+    for value in values:
+        heap.push(int_heap, value)
+        verify(int_heap, 0)
+
+    # random remove item from the heap
+    for i in six.moves.range(n - 1, -1, -1):
+        heap.remove(int_heap, random.randint(0, i))
+        verify(int_heap, 0)
+
```
abhinav: No test for update?

tests/container/test_heap.py
```diff
@@ -0,0 +1,124 @@
...
+
+@pytest.mark.heapfuzz
+@pytest.mark.skipif(True, reason='stress test for the value heap operations')
+def test_heap_fuzz(int_heap):
```
abhinav: How long does this take to run? I'm okay with skipping it by default. We can probably convert it into a pytest-benchmark based benchmark later.

tests/container/test_heap.py
junchaowu: search for test update.

tests/test_peer_heap.py
```diff
@@ -0,0 +1,110 @@
...
+    if child2 < ph.size():
+        assert ph.lt(parent, child1)
+        verify(ph, child1)
+
+    if child2 < ph.size():
+        assert ph.lt(parent, child2)
+        verify(ph, child2)
+
+
+def test_push(peer_heap):
```
abhinav: I wasn't asking to duplicate these tests. test_heap.py is enough to test the heap implementation itself.

For this file, your assumption should be that the heap implementation works as expected. You only need to do a sort of sanity check that PeerHeap's own operations work, and that it does what we say HeapOperation does.

Like, make sure that `pop_peer()` doesn't explode when the list is empty. Stuff like that. Basically, test_heap gives us confidence in the heap implementation and test_peer_heap gives us confidence that PeerHeap knows how to work with that heap implementation without blowing up.

abhinav:
:ok:

Issue #311, opened 2015-11-18
Unexpected error frames should add the exception message to the error frame


prashantv:
If a handler raises an exception, we currently send a generic error frame back:
https://github.com/uber/tchannel-python/blob/master/tchannel/tornado/dispatch.py#L231

We should add the exception message to the error frame so it provides more information to clients.

blampe:
@breerly suggested this originally and I think I was opposed to the idea but I've come around, let's do this. cc @junchaowu

PR #314, opened 2015-11-24
Add more descriptive message to the unexception error message.

junchaowu:
this is going to fix  https://github.com/uber/tchannel-python/issues/311


blampe:
I was thinking of returning the stack but this is a start. :ship:

PR #320, opened 2015-12-03
Add list to track outbound pending req/res

junchaowu:
This will be used for peer selection.

### DIFF
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -143,6 +143,7 @@ class TornadoConnection(object):
 
     @property
     def total_outbound_pendings(self):
+        """Return total amount of outbounding pending req/res."""
         return (len(self._outbound_pending_res) +
                 len(self._outbound_pending_req))
 
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -124,13 +124,13 @@ class TornadoConnection(object):
         self._messages = queues.Queue()
 
         # Map from message ID to futures for responses of outgoing calls.
-        self._out_pending_call = {}
+        self._outbound_pending_call = {}
 
         # Map from message ID to outgoing request that is being sent.
-        self._out_pending_req = {}
+        self._outbound_pending_req = {}
 
         # Map from message ID to outgoing response that is being sent.
-        self._out_pending_res = {}
+        self._outbound_pending_res = {}
 
         # Whether _loop is running. The loop doesn't run until after the
         # handshake has been performed.
@@ -142,8 +142,9 @@ class TornadoConnection(object):
         connection.set_close_callback(self._on_close)
 
     @property
-    def num_out_pendings(self):
-        return len(self._out_pending_res) + len(self._out_pending_req)
+    def total_outbound_pendings(self):
+        return (len(self._outbound_pending_res) +
+                len(self._outbound_pending_req))
 
     def set_close_callback(self, cb):
         """Specify a function to be called when this connection is closed.
@@ -164,13 +165,13 @@ class TornadoConnection(object):
     def _on_close(self):
         self.closed = True
 
-        for message_id, future in self._out_pending_call.iteritems():
+        for message_id, future in self._outbound_pending_call.iteritems():
             future.set_exception(
                 NetworkError(
                     "canceling outstanding request %d" % message_id
                 )
             )
-        self._out_pending_call = {}
+        self._outbound_pending_call = {}
 
         try:
             while True:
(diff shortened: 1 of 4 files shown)


tchannel/tornado/connection.py
```diff
@@ -135,6 +141,10 @@ def __init__(self, connection, tchannel=None, direction=None):
 
         connection.set_close_callback(self._on_close)
 
+    @property
+    def num_out_pendings(self):
```
breerly: suggest `total_out_pendings`

tchannel/tornado/connection.py
```diff
@@ -124,7 +124,13 @@ def __init__(self, connection, tchannel=None, direction=None):
         self._messages = queues.Queue()
 
         # Map from message ID to futures for responses of outgoing calls.
-        self._outstanding = {}
+        self._out_pending_call = {}
```
breerly: suggest renaming these to `_outbound_pending_X`

tchannel/tornado/util.py
```diff
@@ -90,3 +90,12 @@ def go():
     go()
 
     return all_done_future
+
+
+def num_out_pendings(conns):
```
breerly: where does conns come from? seems like this should be a method on a connection group object

tests/tornado/test_connection.py
```diff
@@ -81,3 +82,55 @@ def test_close_callback_is_called():
...
+    yield client.raw(
+        hostport=server.hostport,
+        body='work',
+        endpoint='hello',
+        service='server'
+    )
+
+    assert num_out_pendings(
+        client._dep_tchannel.peers.get(server.hostport).connections
+    ) == 0
```
breerly: This is a bit tricky to test. It might be a symptom of how this code is structured.

You might be able to spy on these outbound pendings list in such a way that you can ask the spy about the state changes that happened - aka assert `num_out_pendings` contained X before it was eventually drained.

tchannel/tornado/connection.py
abhinav: Based on usage it looks like this should be a method on Peer

tchannel/tornado/util.py
abhinav: +1

Based on usage it looks like this should be a method on Peer (which acts as a grouping of connections of sorts).

tchannel/tornado/connection.py
junchaowu: each connection will expose its own one and peers' is the debate below about util method or instance method inside peer.

breerly:
Sure :shipit:

tchannel/tornado/connection.py
```diff
@@ -135,6 +141,11 @@ def __init__(self, connection, tchannel=None, direction=None):
 
         connection.set_close_callback(self._on_close)
 
+    @property
+    def total_outbound_pendings(self):
+        return (len(self._outbound_pending_res) +
```
abhinav: docstring please

tests/tornado/test_connection.py
```diff
@@ -81,3 +82,49 @@ def test_close_callback_is_called():
...
+
+    client = TChannel('client')
+    yield client.raw(
+        hostport=server.hostport,
+        body='work',
+        endpoint='hello',
+        service='server'
+    )
+
+    client_peer = client._dep_tchannel.peers.peers[0]
```
abhinav: Consider using `.peers.get(..)` instead of accessing the list.

tests/tornado/test_connection.py
```diff
@@ -81,3 +82,49 @@ def test_close_callback_is_called():
...
+    assert server_peer.total_outbound_pendings == 0
+
+    class FakeMessageFactory(MessageFactory):
+        def build_raw_message(self, context, args, is_completed=True):
+            assert client_peer.total_outbound_pendings == 1
+            return super(FakeMessageFactory, self).build_raw_message(
+                context, args, is_completed,
+            )
+
+    client_conn = client_peer.connections[0]
```
abhinav: `peer.connect()` will get the connection off the peer. You don't need to access `connections` (we should really hide that behind a `_` anyway).

tests/tornado/test_connection.py
```diff
@@ -81,3 +82,49 @@ def test_close_callback_is_called():
...
+
+    client_peer = client._dep_tchannel.peers.peers[0]
+    server_peer = server._dep_tchannel.peers.peers[0]
+    assert client_peer.total_outbound_pendings == 0
+    assert server_peer.total_outbound_pendings == 0
+
+    class FakeMessageFactory(MessageFactory):
+        def build_raw_message(self, context, args, is_completed=True):
+            assert client_peer.total_outbound_pendings == 1
+            return super(FakeMessageFactory, self).build_raw_message(
```
abhinav: cool so we didn't need to break it across messages after all

abhinav:
lgtm :shipit:

abhinav:
my bad

PR #322, opened 2015-12-04
Add remove_peer method & peer strategy class

junchaowu:
@abhinav @breerly @blampe


tchannel/peer_heap.py
```diff
@@ -80,3 +80,8 @@ def peek_peer(self):
             return None
 
         return self.peers[0]
+
+    def remove_peer(self, peer):
+        """Remove the peer from the heap."""
+        if peer.index < self.size():
```
abhinav: `peer.index` can be -1, right? change this to

    if 0 <= peer.index < self.size():

abhinav: Is it worth verifying that the peer being removed is the same peer? Are we guaranteed that the peer.index will never be out of date?

tests/test_peer_heap.py
```diff
@@ -107,6 +107,18 @@ def test_update(peer_heap, peers):
...
+    for peer in peers:
+        peer_heap.push_peer(peer)
+        verify(peer_heap, 0)
+
+    n = len(peers)
+    for _ in six.moves.range(n):
+        p = peer_heap.peers[random.randint(0, peer_heap.size() - 1)]
+        peer_heap.remove_peer(p)
+        verify(peer_heap, 0)
+
```
abhinav: Add tests for at least the following cases:

- double removal: remove_peer on the same peer twice
- removing a peer from an empty peer heap
- removing a peer that doesn't exist in the heap

blampe: Also please be sure to test this on a dummy service :)

tchannel/peer_heap.py
```diff
@@ -80,3 +80,12 @@ def peek_peer(self):
             return None
 
         return self.peers[0]
+
+    def remove_peer(self, peer):
+        """Remove the peer from the heap.
+        If the peer's index out of range[0, peer'size), the remove function
+        will do nothing.
```
abhinav: Consider changing the second sentence to,

> Unknown peers are ignored.

Range details irrelevant to API.

tchannel/peer_heap.py
```diff
@@ -80,3 +80,12 @@ def peek_peer(self):
...
+        """
+        if 0 <= peer.index < self.size():
+            p = heap.remove(self, peer.index)
+            assert p == peer
```
abhinav: Do we have a `__eq__` definition on Peer? If you just want to do a reference match ("these refer to the same peer in-memory"), do `p is peer`. Otherwise, maybe do `p.hostport == peer.hostport` instead.

tests/test_event.py
```diff
@@ -80,7 +80,7 @@ def test_after_send_error_event_called():
                 service='test',
                 arg1='endpoint',
                 hostport=tchannel.hostport,
-                timeout=0.02,
+                timeout=0.3,
```
abhinav: ?

tests/test_peer_heap.py
```diff
@@ -107,6 +107,45 @@ def test_update(peer_heap, peers):
...
+
+def test_remove_none_exist(peer_heap, peers):
+    for peer in peers:
+        peer_heap.push_peer(peer)
+        verify(peer_heap, 0)
+
+    peer_heap.remove_peer(peers[0])
+
+
```
abhinav: No test for peer mismatch?

tests/test_peer_heap.py
junchaowu: sure. Everything related to heap is landing into peerheap branch first.

tests/test_peer_heap.py
junchaowu: what do you mean peer mismatch?

tests/test_peer_heap.py
```diff
@@ -107,6 +107,45 @@ def test_update(peer_heap, peers):
...
 
+def test_remove(peer_heap, peers):
+    for peer in peers:
+        peer_heap.push_peer(peer)
+        verify(peer_heap, 0)
+
+    n = len(peers)
+    for _ in six.moves.range(n):
+        p = peer_heap.peers[random.randint(0, peer_heap.size() - 1)]
+        peer_heap.remove_peer(p)
```
abhinav: Also, is there a way to verify here that the peer is no longer in the heap?

tchannel/peer_heap.py
abhinav: Also, this assertion is coming after you've already removed the peer. So if it fails, you'd have removed the wrong peer anyway.

tests/test_peer_heap.py
junchaowu: Added the test.

tests/test_peer_heap.py
```diff
@@ -107,6 +107,57 @@ def test_update(peer_heap, peers):
...
+
+def test_remove_mismatch(peer_heap, peers):
+    for peer in peers:
+        peer_heap.push_peer(peer)
+        verify(peer_heap, 0)
+
+    # create a fake peer with duplicated index.
+    fake_peer = mock_peer()
+    fake_peer.index = 1
+    with pytest.raises(AssertionError):
```
abhinav: Can't you reproduce this "legally" with this:

    heap.push(peer)
    heap.remove(peer)
    heap.push(other_peer_with_same_score)
    heap.remove(peer)

abhinav: Actually, keep this test and add a test for that too.

tchannel/peer_heap.py
abhinav: ping

tests/test_peer_heap.py
junchaowu: this is not valid case. same score doesn't mean same index. Remove peer is based on the index.
That's why I write it by changing the index.

tchannel/peer_heap.py
abhinav: also, please add a reasonable error message here.

tests/test_peer_heap.py
abhinav: Don't you need to call `verify_peer_not_in_heap` here too?

tests/test_peer_heap.py
```diff
@@ -107,6 +107,63 @@ def test_update(peer_heap, peers):
...
+        peer_heap.remove_peer(fake_peer)
```
abhinav: once you add an assertion error message, can you check here that it matches?

tests/test_event.py
junchaowu: the timeout set too small causing Timeout error failing the test.

tests/test_peer_heap.py
junchaowu: Added

tests/test_peer_heap.py
abhinav: fair enough. making it end up at the same index would be hard with just the same score. what if you remove the peer at index 0 and the new peer has the lowest score so it ends up at index 0 too?

tests/test_peer_heap.py
junchaowu: I already has this assertion. I don't need to add assert peer matches. If peer mismatch, it will raise the assertion

abhinav: No, you need to check that the error you expected was raised. You're not using a unique exception type for this error. You're using a very general exception type. Anything else could raise assertion error too. You need to match on the error message to make sure that it's the assertion error you expected.

tchannel/peer_heap.py
abhinav: ping. github is hiding this.

tests/test_peer_heap.py
```diff
@@ -107,6 +107,64 @@ def test_update(peer_heap, peers):
...
+def test_remove_mismatch(peer_heap, peers):
+    for peer in peers:
+        peer_heap.push_peer(peer)
+        verify(peer_heap, 0)
+
+    # create a fake peer with duplicated index.
+    fake_peer = mock_peer()
+    fake_peer.index = 1
+    with pytest.raises(AssertionError):
+        assert fake_peer is peer_heap.remove_peer(fake_peer)
```
abhinav: What's this? This doesn't help the test. In fact it introduces the possibility that remove_peer behaves incorrectly and this test still passes.

This is not what I meant when I said you need to test that the correct assertion error was raised. You need to assert that the `AssertionError` was thrown with `pytest.raises` and then check that the error message matches what you expect it to be. See https://pytest.org/latest/assert.html#assertions-about-expected-exceptions

tchannel/peer_heap.py
```diff
@@ -80,3 +80,15 @@ def peek_peer(self):
...
+        """Remove the peer from the heap.
+        Note: Unknown peers are ignored.
+
+        Return: removed peer if peer exists, Otherwise return None.
+        """
+        if 0 <= peer.index < self.size():
+            assert (peer is self.peers[peer.index],
+                    "peer doesn't exist in the heap")
+            p = heap.remove(self, peer.index)
+            return p
```
abhinav: I know you said that unknown peer index is a no-op but there seems to be a discrepancy in behavior here.

- If your peer is unknown AND it has an index outside our range, we'll do nothing.
- If your peer is unknown but its index is within our range, we'll raise an exception.

I think we should raise an exception in both cases. In the second case, you can raise an IndexError for the out-of-range case (the one where it's a no-op).

tests/test_event.py
blampe: Are you sure we didn't accidentally make something a lot slower?

junchaowu: 0.02 timeout is little bit aggressive and make the test flaky. Make it a little bit bigger.

abhinav:
lgtm :shipit: 

That 10x increase in the timeout seems concerning. Profile that when you get a chance.

PR #323, opened 2015-12-05
Add support for registering fallback endpoints.

abhinav:
Expose request, timeout, and trace in the Request object inside handlers.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,8 @@ Changes by Version
 - ``tchannel.TChannel.FALLBACK`` may now be used to register fallback endpoints
   which are called for requests with unrecognized endpoints. For more
   information, see :ref:`fallback-endpoint`
-- Expose ``timeout``, ``service``, and ``trace`` attributes on ``Request``
-  objects inside endpoint handlers.
+- Expose ``timeout`` and ``service`` attributes on ``Request`` objects inside
+  endpoint handlers.
 
 
 0.20.2 (2015-11-25)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,11 @@ Changes by Version
 -------------------
 
 - Add support for ``rd`` transport header.
+- ``tchannel.TChannel.FALLBACK`` may now be used to register fallback endpoints
+  which are called for requests with unrecognized endpoints. For more
+  information, see :ref:`fallback-endpoint`
+- Expose ``timeout``, ``service``, and ``trace`` attributes on ``Request``
+  objects inside endpoint handlers.
 
 
 0.20.2 (2015-11-25)
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -8,8 +8,8 @@ Can I register an endpoint that accepts all requests?
 
 The fallback endpoint is the endpoint called when an unrecognized request is
 received. By default, the fallback endpoint simply returns a
-``BadRequestError`` to the caller. This behavior may be overriden by
-registering a raw endpoint with ``TChannel.FALLBACK``.
+``BadRequestError`` to the caller. This behavior may be changed by
+registering an endpoint with ``TChannel.FALLBACK``.
 
 .. code-block:: python
 
@@ -17,7 +17,7 @@ registering a raw endpoint with ``TChannel.FALLBACK``.
 
     server = TChannel(name='myservice')
 
-    @server.register('raw', TChannel.FALLBACK)
+    @server.register(TChannel.FALLBACK)
     def handler(request):
         # ...
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,0 +1,27 @@
+FAQ
+===
+
+.. _fallback-endpoint:
+
+Can I register an endpoint that accepts all requests?
+-----------------------------------------------------
+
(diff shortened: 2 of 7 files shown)


abhinav:
@blampe @breerly @junchaowu

docs/faq.rst
```diff
@@ -0,0 +1,27 @@
...
+``BadRequestError`` to the caller. This behavior may be overriden by
+registering a raw endpoint with ``TChannel.FALLBACK``.
+
+.. code-block:: python
+
+    from tchannel import TChannel
+
+    server = TChannel(name='myservice')
+
+    @server.register('raw', TChannel.FALLBACK)
```
blampe: Does a fallback need to registered for every scheme?

tchannel/request.py
```diff
@@ -47,6 +47,16 @@ class Request(object):
         The most useful piece of information here is probably
         ``request.transport.caller_name``, which is the identity of the
         application that created this request.
+
+    :ivar service:
+        Name of the service being called.
```
blampe: How would this differ from the name you advertised as? Is this related to rd transport header?

tchannel/request.py
```diff
@@ -47,6 +47,16 @@ class Request(object):
...
+
+    :ivar timeout:
+        Amount of time (in seconds) within which this request is expected to
+        finish.
+
+    :ivar trace:
```
blampe: Doesn't this already live on the request context? Seems more extensible to attach the whole request context object to the request.

tchannel/request.py
```diff
@@ -54,8 +64,11 @@ class Request(object):
     __slots__ = (
         'body',
         'headers',
+        'service',
```
junchaowu: is this just for migration information from old request to new request.

docs/faq.rst
abhinav: No. Raw counts as all schemes. Basically, the scheme equality check is skipped if the handler is a fallback.

tchannel/request.py
abhinav: This is the name of the target service that the caller provided. Most of the time it will be "this" service but when you're a forwarding service, it'll be whatever you're supposed to forward this to.

tests/test_tchannel.py
```diff
@@ -376,3 +376,99 @@ def test_listen_duplicate_ports():
...
+        response = yield proxy_server_client.call(
+            scheme=request.transport.scheme,
+            service=request.service,
+            arg1=request.endpoint,
+            arg2=request.headers,
+            arg3=request.body,
+            timeout=request.timeout / 2,
+            retry_on=request.transport.retry_flags,
+            retry_limit=0,
+            shard_key=request.transport.shard_key,
```
blampe: rd instead of sk?

tchannel/request.py
blampe: Ah yeah, maybe make a note that it's most relevant to proxies.

tchannel/request.py
abhinav: Yeah, it does. Currently, using this will override the context's trace info and use the one from the original request.

You think we should just let the default behavior do its thing? That definitely seems more correct because that way the forwarding service itself will show in the trace too.

abhinav: Fixed. Tracing information is not exposed anymore.

docs/faq.rst
abhinav: This does seem a bit janky. I could add some hacks in place to say scheme isn't required if you're using fallback but that'll complicate an already complicated API.

tchannel/request.py
abhinav: Added note

tchannel/request.py
abhinav: No. This is the service name the caller gave in the call. If the service is a proxy that will just forward requests, then the service name of the target service will be here.

tests/test_tchannel.py
abhinav: I don't understand

docs/faq.rst
abhinav: Actually I figured out a less dirty way to make this happen. Turns out that the old fallback registration code already expected scheme to be absent. Fix incoming.

tests/test_tchannel.py
abhinav: Done #325.

abhinav: Okay, so `rd` is going to 0.21. This is going to master. I'm gonna hold off on adding it here.

blampe:
LGTM after you fix transport header.

tests/test_tchannel.py
abhinav: nvm, added to master. #326. Will update this PR once that lands.

CHANGES.rst
```diff
@@ -4,7 +4,11 @@ Changes by Version
 0.20.3 (unreleased)
```
breerly: Seems like a minor release yeah?

abhinav: true. we keep bumping that other release :(

tchannel/request.py
breerly: Wonder if `caller` is a better name. @blampe whatchutink?

breerly: Aha - nm.

breerly:
:shipit:

PR #327, opened 2015-12-07
Add peer score update event

junchaowu:
@breerly @abhinav @blampe


benchmarks/test_peer_selection.py
```diff
@@ -39,6 +39,7 @@ class FakeConnection(object):
     def __init__(self, hostport):
         self.hostport = hostport
         self.closed = False
+        self.direction = object()
```
blampe: ?

tchannel/peer_heap.py
```diff
@@ -31,11 +33,26 @@ class PeerHeap(HeapOperation):
...
     def size(self):
         return len(self.peers)
 
     def lt(self, i, j):
+        """Compare the priority of two peers.
+
+        Primary comparator will be the score of each peer. If the score is
+        same then use FIFO policy. The ``order`` attribute of the peer tracks
+        the heap push order of the peer. This help solve the imbalance
+        problem caused by randomization when deal with same score situation.
```
abhinav: Good work on adding the documentation but the information on the specifics of how the peer heap behaves should be in the class doc, not individual method docs.

tchannel/peer_heap.py
```diff
@@ -67,7 +84,28 @@ def pop_peer(self):
...
+        same list of hyperbahn nodes, the peers that they pick to talk to
+        hyperbahn will be same. In that situation, it will cause the hyperbahn
+        node received very imbalance requests from tchannel clients.
+
+        All in all, it will keep certain level randomization but at the same
+        time make the peer score not deterministic among different tchannel
+        instance. For
+        """
+        self.order += 1
+        peer.order = self.order + random.randint(0, self.size())
```
abhinav: I'm confused. How does this solve problem 2? `peer.order` doesn't track the exact insertion order but it's still monotonically increasing with respect to peer insertion. So if we `push(A)`, and then `push(B)`, we will always have `A.order < B.order`, which means that score collisions will always resolve in a deterministic way?

tchannel/peer_heap.py
```diff
@@ -31,11 +33,26 @@ class PeerHeap(HeapOperation):
...
+        self.order = 0
 
     def size(self):
         return len(self.peers)
 
     def lt(self, i, j):
+        """Compare the priority of two peers.
+
+        Primary comparator will be the score of each peer. If the score is
+        same then use FIFO policy. The ``order`` attribute of the peer tracks
```
abhinav: Not exactly FIFO since you're trying to avoid that 100% deterministic behavior.

tchannel/peer_strategy.py
```diff
@@ -30,3 +32,23 @@ def get_score(self, peer):
...
+        In our peer selection strategy, the largest number has least priority
+        in the heap.
+
+        If the peer has incoming connections, we will return number of outbound
+        pending requests and responses.
+
+        :param peer: instance of `tchannel.tornado.peer.Peer`
+        :return: score of the peer
+        """
+        if len(peer.incoming_connections) == 0:
```
abhinav: if not peer.incoming_connections:
         # ...

benchmarks/test_peer_selection.py
junchaowu: It is because in the "self.score_calculator.get_score(peer)" will need to get the incoming connections which requires the direction attributes. So I added it here.

abhinav: Can you give this a valid value? INCOMING/OUTGOING?

tests/tornado/test_connection.py
```diff
@@ -128,3 +129,52 @@ def build_raw_message(self, context, args, is_completed=True):
...
+        return 'hi'
+
+    client = TChannel('client')
+    count = [0]
+
+    def test_cb(peer):
+        count[0] += 1
+
+    client._dep_tchannel.peers.get(
+        server.hostport)._on_conn_change_cb = test_cb
```
abhinav: No. Test these directly on the Peer or PeerGroup object. Private attribute access in a test is generally an indicator that you need to change the test.

tests/tornado/test_connection.py
```diff
@@ -128,3 +129,52 @@ def build_raw_message(self, context, args, is_completed=True):
...
+
+    # 1: connection built, 1: sending request, 1: finish sending request
+    assert count[0] == 3
+
+
+@pytest.mark.gen_test
+def test_both_connection_change_callback():
+    client = TChannel('client')
+
+    with mock.patch.object(Peer, '_on_conn_change') as mock_conn_change:
```
abhinav: Again, test the Peer objects directly.

tchannel/peer_heap.py
junchaowu: the problem 2 is solved in docstring below.

abhinav: I don't know which docstring you're referring to. This is the last change in this file. There's nothing below it.

tests/tornado/test_connection.py
junchaowu: what do you mean? If you think access private in test not proper, I can add method to set callback. Also can you give example on "directly on the Peer or PeerGroup object" ?

tchannel/peer_heap.py
junchaowu: Oh it is the code "random.randint(0, self.size())" will make that not deterministic. Cause the order is not same all the time because of random part.

tchannel/peer_heap.py
junchaowu: Right, I should check the FIFIO to based on sub field 'order'.

benchmarks/test_peer_selection.py
junchaowu: It won't change the behavior actually.

abhinav: Yeah, then do set it to INCOMING/OUTGOING. Makes it obvious what you're trying to do there.

tchannel/peer_heap.py
abhinav: So, if you write a test:

    peers = [p1, p2, p3, p4]  # where p{1,2,3,4} have the same score
    for p in peers:
        heap.push(p)

    assert heap.peek_peer() == p1  # this will always be true

This test will always pass even though this is something you're trying to avoid.

It will always pass because even though orders for the four peers aren't 1, 2, 3, 4, `p1.order` is always smaller than order for p2, p3, or p4.

tchannel/peer_heap.py
```diff
@@ -31,11 +33,26 @@ class PeerHeap(HeapOperation):
...
+        the heap push order of the peer. This help solve the imbalance
+        problem caused by randomization when deal with same score situation.
+
+        :param i: ith peer
+        :param j: jth peer
+        :return: True or False
+        """
+        if self.peers[i].score == self.peers[j].score:
+            return self.peers[i].order < self.peers[i].order
```
abhinav: The right one should be j

tests/tornado/test_connection.py
abhinav: I mean that these tests should be instantiating the Peer class and not accessing it via private variables in the TChannel class. Test the contract of the Peer class. It has the appropriate methods.

You need a series of different tests:

- Tests on the Connection class that assert that the callbacks for the different cases are called as expected. These tests should instantiate the connection class, not the TChannel class.
- Tests on the Peer class that assert that the callback is called as expected. Again, these tests should instantiate the Peer class, call methods on Peer, etc.

tchannel/peer_heap.py
junchaowu: As the peer selection goes, the p1's order = current_order + random(sizeof(heap)) 
the second part value range from 0 to size of heap. I don't understand why you think it's not random.

tchannel/peer_heap.py
junchaowu: nice catch.

tchannel/peer_heap.py
abhinav: Oh true. My bad

tchannel/peer_heap.py
```diff
@@ -67,7 +84,28 @@ def pop_peer(self):
...
+
+        1. 'dead peer' caused by random problem. If use fully random value, and
+        one peer is very unlock to get the worst value, as the peer selection
+        keeps working, that peer may not be selected anymore.
+
+        2. deterministic problem. If we just use push order, then all the
+        tchannel instance will follow the same pattern. If they all get the
+        same list of hyperbahn nodes, the peers that they pick to talk to
+        hyperbahn will be same. In that situation, it will cause the hyperbahn
+        node received very imbalance requests from tchannel clients.
```
abhinav: Change to:

> - Dead peers: If the order is completely random, then an unlucky peer with a very bad assigned order may never get selected.
> - Determinism: If the insertion order is used as-is, then all TChannel instances would follow the same selection pattern, causing load imbalance. For example if they get the same static list of peers, they will all pick the first one, then the second one, and so on, cycling between which host gets overloaded with requests.

Also, same concern as before: This behavioral description (and how PeerHeap solves it with order) belongs in the class docstring, not the method docstring. Like,

    class MyClass(object):
        """MyClass does X.

        It does Z to solve the following problems:
        - ...
        """

tchannel/peer_strategy.py
```diff
@@ -30,3 +32,23 @@ def get_score(self, peer):
...
+        In our peer selection strategy, the largest number has least priority
+        in the heap.
+
+        If the peer has incoming connections, we will return number of outbound
+        pending requests and responses.
+
+        :param peer: instance of `tchannel.tornado.peer.Peer`
+        :return: score of the peer
+        """
+        if not peer.incoming_connections:
```
abhinav: Do we need to make sure that the peer is not 0.0.0.0:0. I don't think we're checking that right now and that could cause problems with tcurl.

tchannel/peer_strategy.py
```diff
@@ -30,3 +32,24 @@ def get_score(self, peer):
...
+        In our peer selection strategy, the largest number has least priority
+        in the heap.
+
+        If the peer has incoming connections, we will return number of outbound
+        pending requests and responses.
+
+        :param peer: instance of `tchannel.tornado.peer.Peer`
+        :return: score of the peer
+        """
+        if (peer.host == '0.0.0.0' or int(peer.port) == 0 or
```
abhinav: This can be collapsed into `peer.is_ephemeral`.

tchannel/peer_strategy.py
```diff
@@ -30,3 +32,24 @@ def get_score(self, peer):
...
+                not peer.incoming_connections):
```
abhinav: Skipping peers that don't have any incoming connections seems risky. I'm imagining a scenario where you have one incoming connection and N outgoing connections. This will cause all the load to be sent to that one incoming connection. We shouldn't tie the scoring behavior to how Hyperbahn works right now.

tchannel/tornado/peer.py
```diff
@@ -583,7 +628,9 @@ def get(self, hostport):
         """
         assert hostport, "hostport is required"
         if hostport not in self._peers:
-            self._peers[hostport] = self.peer_class(self.tchannel, hostport)
+            self.add(hostport)
+            return self.get(hostport)
```
abhinav: Unnecessary recursive call. Change this to

    if hostport not in self._peers:
        self.add(hostport)
    return self._peers[hostport]

tests/test_peer_strategy.py
```diff
@@ -0,0 +1,62 @@
...
+    assert sys.maxint == calculator.get_score(peer)
+
+    connection.direction = INCOMING
+    assert sys.maxint != calculator.get_score(peer)
+
+    peer.port = 0
+    peer.host = '1.1.1.1'
+    calculator = PreferIncomingCalculator()
+    assert sys.maxint == calculator.get_score(peer)
+
```
abhinav: Please split these into different test functions with descriptive names

tests/tornado/test_connection.py
abhinav: You can remove these tests now that this functionality is being correctly tested on the Peer object.

Might be worth testing that PeerGroup behaves correctly too. Something like:

    peer = peer_group.choose()
    # start request to peer
    other_peer = peer_group.choose()
    assert other_peer is not peer

tchannel/peer_strategy.py
junchaowu: I totally agree that more sophisticated way to can put in the strategy. For the init change, let's start from simple. We can always add new strategy later , the design supports new strategy.

abhinav: I'm all for incremental improvements but this strategy just has a very bad behavior for the outgoing connections case. I would drop the incoming connections requirement for now until we can figure out a smarter strategy there.

abhinav:
Oh, I see GH was nice enough to hide my comment. Here it is again.

In response to https://github.com/uber/tchannel-python/pull/327#discussion_r47574844

> I'm all for incremental improvements but this strategy just has a very bad behavior for the outgoing connections case. I would drop the incoming connections requirement for now until we can figure out a smarter strategy there.

junchaowu:
That's how it works on Go, node, and python. If you would like a better, talk to @kriskowal @ShanniLi

blampe:
@abhinav is right that this will prevent us from "spilling over" to unused peers in the case of an overloaded connection. I agree that we shouldn't assume anything about HB's behavior right now as our routing strategy will likely be changing.

abhinav:
Also, in addition to that, how does this behave w.r.t. unconnected peers vs connected peers. If I'm not misreading this, unconnected peers will always have a better score because their pending count will be zero. You need to take that into account when scoring. You can't completely skip unconnected peers either because again, we need to be able to establish new connections if we have none.

junchaowu:
@abhinav @blampe  I think we have discussed the peer score strategy offline. Do you have more comments on  this diff?

tchannel/peer_strategy.py
```diff
@@ -30,3 +32,28 @@ def get_score(self, peer):
...
+        :param peer: instance of `tchannel.tornado.peer.Peer`
+        :return: score of the peer
+        """
+        if peer.is_ephemeral or not peer.connections:
+            return self.TIERS[0]
+
+        if not peer.incoming_connections:
+            return self.TIERS[1] + peer.total_outbound_pendings
+
+        return self.TIERS[2] + peer.total_outbound_pendings
```
abhinav: add a small comment about what TIERS is accomplishing

abhinav:
minor comment and then :shipit:

PR #333, opened 2015-12-07
Add test util to test request directly to the endpoint

junchaowu:
It is reopened for 
 https://github.com/uber/tchannel-python/pull/307

@abhinav @blampe @breerly

### DIFF
--- a/tests/testing/test_dispatch.py
+++ b/tests/testing/test_dispatch.py
@@ -1,0 +1,39 @@
+from __future__ import absolute_import
+
+import pytest
+
+from tchannel import TChannel, Request
+from tchannel.errors import BadRequestError
+from tchannel.testing.dispatch import handle
+
+
+@pytest.fixture
+def server():
+    tchannel = TChannel('test_server')
+
+    @tchannel.raw.register
+    def hello(request):
+        return request.body + " world"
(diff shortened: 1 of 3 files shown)


tests/testing/test_dispatch.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import absolute_import
+
+import pytest
+
+from tchannel import TChannel, Request
+from tchannel.errors import BadRequestError
+from tchannel.testing.dispatch import handle
```
blampe: Same feedback as before, what's preventing us from calling the method directly?

junchaowu: because if we call handler directly, the return value are not guaranteed. It could be a string or a float to treat as response body while it also could be a response object. This method will convert all possibilities into one type of result which is tchannel response. As client, it would like to check the response, service own can double check their return value will be converted into right response object.

breerly:
How does this fit in with #337

junchaowu:
the problem is right now endpoint doesn't always return a response object. It only returns a body sometimes.

breerly:
@abhinav lets close this - we should make the endpoints callable in 1.0 instead

PR #335, opened 2015-12-07
disable retry on tchannel zipkin submit

junchaowu:
@blampe @abhinav @breerly

### DIFF
--- a/tchannel/zipkin/tracers.py
+++ b/tchannel/zipkin/tracers.py
@@ -38,6 +38,7 @@ import logging
 import sys
 from collections import defaultdict
 
+from tchannel import retry
 from . import glossary
 from .tcollector import TCollector
 from .formatters import json_formatter
@@ -142,7 +143,7 @@ class TChannelZipkinTracer(object):
 
         def submit_callback(f):
             if f.exception():
-                log.error(
+                log.info(
                     'Fail to submit zipkin trace',
                     exc_info=f.exc_info()
                 )
@@ -152,6 +153,9 @@ class TChannelZipkinTracer(object):
             f = self._tchannel.thrift(
                 TCollector.submit(thrift_formatter(trace, annotations)),
                 shard_key=i64_to_base64(trace.trace_id),
+                retry_on=retry.NEVER,
+                retry_limit=0,
+                trace=False,
             )
             f.add_done_callback(submit_callback)
             fus.append(f)
--- a/tchannel/zipkin/tracers.py
+++ b/tchannel/zipkin/tracers.py
@@ -144,7 +144,7 @@ class TChannelZipkinTracer(object):
         def submit_callback(f):
             if f.exception():
                 log.info(
-                    'Fail to submit zipkin trace',
+                    'Failed to submit zipkin trace',
                     exc_info=f.exc_info()
                 )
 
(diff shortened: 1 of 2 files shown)


tchannel/zipkin/tracers.py
```diff
@@ -142,7 +143,7 @@ def record(self, traces):
 
         def submit_callback(f):
             if f.exception():
-                log.error(
+                log.info(
                     'Fail to submit zipkin trace',
```
blampe: failed

blampe:
YES PLZ

Worth testing?

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -136,3 +137,44 @@ def test_tcollector_submit(trace_server):
...
+    zipkin_server1.thrift.register(TCollector, handler=submit)
+
+    @zipkin_server1.raw.register
+    def hello(request):
+        return 'hello'
+
+    zipkin_server1.listen()
+
+    client = TChannel('client')
+    client._dep_tchannel.peers.add(zipkin_server.hostport)
```
abhinav: Use known peers argument on tchannel instead

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -136,3 +137,44 @@ def test_tcollector_submit(trace_server):
...
+    zipkin_server = TChannel('zipkin')
+    zipkin_server.thrift.register(TCollector, handler=submit)
+    zipkin_server.listen()
+
+    zipkin_server1 = TChannel('server')
+    zipkin_server1.thrift.register(TCollector, handler=submit)
+
+    @zipkin_server1.raw.register
+    def hello(request):
+        return 'hello'
```
abhinav: Not a zipkin sever. That implies this implements the tcollector interface. Please call it something else.

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -136,3 +137,44 @@ def test_tcollector_submit(trace_server):
     results = yield TChannelZipkinTracer(tchannel).record([(trace, anns)])
 
     assert results[0].body.ok is True
+
+
+@pytest.mark.gen_test
+def test_tcollector_submit_never_retry():
+
+    def submit(*args, **kwargs):
```
abhinav: Single 'request' arg

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -136,3 +137,44 @@ def test_tcollector_submit(trace_server):
...
+    client = TChannel('client')
+    client._dep_tchannel.peers.add(zipkin_server.hostport)
+
+    client.hooks.register(ZipkinTraceHook(tchannel=client))
+    yield client.raw(
+        service='server',
+        endpoint='hello',
+        hostport=zipkin_server1.hostport,
+        body='body',
+        trace=True,
```
abhinav: Do we really need to explicitly set trace to true?

tests/integration/trace/test_zipkin_trace.py
junchaowu: this one also implement the tcollector interfactor. The reason is because we can't control which peer to be selected for zipkin tcollector submit. If the submit request goest to the 'hello' service host, it will generate the badrequesterro, which is not retryable.

abhinav: Ugh... Okay.

tests/integration/trace/test_zipkin_trace.py
junchaowu: yes , it is the traceflag.

abhinav: I thought this was automatically enabled if you were using the zipkin hook? I guess my question should've been: do users have to specify `trace=True` for Zipkin tracing to happen?

abhinav:
:+1:

PR #336, opened 2015-12-08
Add sampling logic into the zipkin trace

junchaowu:
@breerly @blampe @abhinav

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Changes by Version
   information, see :ref:`fallback-endpoint`
 - Expose ``timeout`` and ``service`` attributes on ``Request`` objects inside
   endpoint handlers.
+- Add zipkin trace sampling logic and disable the retry for all zipkin trace submit.
 
 
 0.20.2 (2015-11-25)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,8 @@ Changes by Version
   information, see :ref:`fallback-endpoint`
 - Expose ``timeout`` and ``service`` attributes on ``Request`` objects inside
   endpoint handlers.
-- Add zipkin trace sampling logic and disable the retry for all zipkin trace submit.
+- Add support for zipkin trace sampling.
+- Disable the retry for all zipkin trace submit.
 
 
 0.20.2 (2015-11-25)
--- a/tchannel/zipkin/trace.py
+++ b/tchannel/zipkin/trace.py
@@ -40,12 +40,9 @@ import random
 def _uniq_id():
     """Create a random 64-bit signed integer.
 
-    Note: By experimentation Zipkin has trouble recording traces with ids
-    larger than (2 ** 56) - 1.
-
     :rtype: int
     """
-    return random.randint(0, (2 ** 56) - 1)
+    return random.getrandbits(63)
 
 
 class Trace(object):
--- a/tchannel/zipkin/zipkin_trace.py
+++ b/tchannel/zipkin/zipkin_trace.py
@@ -29,7 +29,7 @@ from tchannel.zipkin.tracers import TChannelZipkinTracer
 class ZipkinTraceHook(EventHook):
     """generate zipkin-style span for tracing"""
 
-    DEFAULT_RATE = 0.01
+    DEFAULT_RATE = 1.0
 
     def __init__(self, tchannel=None, dst=None, sample_rate=None):
         """Log zipkin style trace.
@@ -42,7 +42,7 @@ class ZipkinTraceHook(EventHook):
             The sample_rate determines the probability that the trace span
             been sampled.
             The rate of sampling is in the range [0, 1] with 0.01 precision.
-            By default it takes 1% sampling.
+            By default it takes 100% sampling.
         """
 
         if tchannel:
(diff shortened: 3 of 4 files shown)


tchannel/zipkin/trace.py
```diff
@@ -40,12 +40,9 @@
 def _uniq_id():
     """Create a random 64-bit signed integer.
 
-    Note: By experimentation Zipkin has trouble recording traces with ids
-    larger than (2 ** 56) - 1.
-
     :rtype: int
     """
-    return random.randint(0, (2 ** 56) - 1)
+    return random.getrandbits(63)
```
breerly: Oh nice, didn't know this func existed :)

tchannel/zipkin/zipkin_trace.py
```diff
@@ -29,13 +29,18 @@
 class ZipkinTraceHook(EventHook):
     """generate zipkin-style span for tracing"""
 
-    def __init__(self, tchannel=None, dst=None):
+    DEFAULT_RATE = 0.01
```
breerly: Default 1%?

breerly: Don't we only want to be sampling from the edge?

junchaowu: Fix the issue. Should only sampling on the edge.

tchannel/zipkin/trace.py
abhinav:
```
-    Note: By experimentation Zipkin has trouble recording traces with ids
-    larger than (2 ** 56) - 1.
```

Is that no longer true? If it is, maybe this should be 56, not 63?

tchannel/zipkin/zipkin_trace.py
```diff
@@ -29,13 +29,17 @@
...
+
+    def __init__(self, tchannel=None, dst=None, sample_rate=None):
         """Log zipkin style trace.
 
         :param tchannel:
             The tchannel instance to send zipkin trace spans
         :param dst:
             The destination to output trace information
+        :param sample_rate:
+            The rate of sampling is in the range (0, 1] with 0.01 precision.
```
abhinav: Documentation should mention what the field is/how it affects behavior, not just the valid values.

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -176,3 +178,17 @@ def hello(request):
...
+def test_zipkin_trace_sampling(trace_server):
+    run_times = 100000
+    sample_rate = 1.0 - random.random()
+    hook = ZipkinTraceHook(sample_rate=sample_rate)
+    count = 0
+    for _ in range(run_times):
+        if hook._lucky(_uniq_id()):
+            count += 1
+
+    assert 0.9 * run_times * sample_rate <= count
```
abhinav: This test in inadequate. Replace this with:

- A test that ensures that no traces are emitted if rate is 0 even if trace is set to True
- A test that mocks `_lucky` to emit True for a specific case and checks that the trace is emitted. (We want to avoid mocking internal APIs usually but since in this case we can't control randomness cleanly, we have to do this.)

tchannel/zipkin/trace.py
junchaowu: that's the old note added by Bryce which is no longer valid.

tchannel/zipkin/zipkin_trace.py
junchaowu: sample_rate is clear enough. Won't fix.

tests/integration/trace/test_zipkin_trace.py
junchaowu: Yeah. I added the test for the first case.
For the second situation, the previous two tests that I set the sample_rate to 1, equals to make the lucky return True and verify the call been made.

tchannel/zipkin/trace.py
prashantv: Why 63 instead of 64?

tests/integration/trace/test_zipkin_trace.py
abhinav: My issue is with this specific test trying to measure randomness. I'm suggesting deleting this test in favor of something like,

    hook = ZipkinTraceHook(...)
    with mock(hook, '_lucky') as mock_lucky:
        mock_lucky.return_value = True
        request = Request()
        hook.before_send_request(request)
        assert ...

Usually we'd want to avoid something like that because of mocking `_lucky` but in this case there's no way for us to reasonably control the randomness without relying on more specific details of how `_lucky` is implemented.

abhinav:
Update changelog please

tests/integration/trace/test_zipkin_trace.py
junchaowu: this test is for testing the sampling algorithm. Not test the trace call.

abhinav: Okay I can buy that

abhinav:
Changelog and then :shipit:

CHANGES.rst
```diff
@@ -10,6 +10,7 @@ Changes by Version
   information, see :ref:`fallback-endpoint`
 - Expose ``timeout`` and ``service`` attributes on ``Request`` objects inside
   endpoint handlers.
+- Add zipkin trace sampling logic and disable the retry for all zipkin trace submit.
```
abhinav: These are two separate points

> - Disable retrying of Zipkin requests.
> - Add support for sampling a subset of requests when submitting Zipkin traces.

PR #337, opened 2015-12-09
Change tchannel.thrift.register to return the unchanged handler.

abhinav:
This is a backport of #302 to 0.21.

### DIFF
--- a/tests/schemes/test_thrift.py
+++ b/tests/schemes/test_thrift.py
@@ -23,6 +23,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import mock
 import pytest
 from tornado import gen
 from tornado import concurrent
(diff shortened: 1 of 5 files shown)


abhinav:
@blampe @breerly @junchaowu

tests/schemes/test_thrift.py
```diff
@@ -1162,3 +1163,80 @@ def testException(request):
...
+def test_void_with_headers_call_directly(server, ThriftTest):
+
+    @server.thrift.register(ThriftTest)
+    def testVoid(request):
+        assert request.headers == {'req': 'header'}
+        return Response(headers={'resp': 'header'})
+
+    resp = testVoid(Request(headers={'req': 'header'}))
+    assert resp.headers == {'resp': 'header'}
+    assert resp.body is None
```
breerly: This is awesome.

tests/schemes/test_thrift.py
```diff
@@ -1162,3 +1163,80 @@ def testException(request):
...
+    def testStruct(request):
+        assert request.headers == {'req': 'header'}
+        assert request.body.thing.string_thing == 'req string'
+
+        return Response(
+            server_ttypes.Xtruct(string_thing="resp string"),
+            headers={'resp': 'header'},
+        )
+
+    resp = testStruct(
```
breerly: so stoked on this.

tests/schemes/test_thrift.py
```diff
@@ -1162,3 +1163,80 @@ def testException(request):
...
 
     assert 1 == res.status_code
+
+
+##############################################################################
+# Calling registered functions directly should be equivalent to calling them
+# as if they weren't registered at all
+
+
+def body(**kwargs):
```
breerly: guessing we'll have to do work to make the request easier to inject into the endpoint?

breerly:
Pumped. :shipit:

tests/schemes/test_thrift.py
abhinav: Yes, once we settle on an API to let users construct the args structs for different service methods.

I was thinking something like, but let's discuss that elsewhere.

```python
@tchannel.thrift.register
def doFoo(request):
    # ...

def test_do_foo():
    # ...
    doFoo(Request(body=doFoo.args(1, 2, 3)))
```

PR #338, opened 2015-12-09
Fix the bug to generator signed trace_id

junchaowu:
@blampe @abhinav @breerly 
In the rw.py, we always used 'q' for struct.pack or unpack. It treats all numbers as signed long long. I double checked the our protocol, all the 64bits data are not signed. So change it to 'Q'

### DIFF
--- a/tchannel/rw.py
+++ b/tchannel/rw.py
@@ -347,7 +347,7 @@ class NumberReadWriter(ReadWriter):
         1: '>B',
         2: '>H',
         4: '>I',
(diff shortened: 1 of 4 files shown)


breerly:
test?

abhinav:
:ok: 

Worth noting that `struct.unpack('>Q', -1)` fails with an exception while `>q` doesn't.

junchaowu:
I believe all our existing tests have covered the usage of struct.pack 'Q'

PR #339, opened 2015-12-10
Fix JSON argscheme header encoding

abhinav:
Pending confirmation that it was wrong

### DIFF
--- a/tchannel/serializer/json.py
+++ b/tchannel/serializer/json.py
@@ -23,14 +23,27 @@ from __future__ import absolute_import
 import json
 
 from tchannel.schemes import JSON
-from tchannel.serializer.thrift import ThriftSerializer
 
 
-class JsonSerializer(ThriftSerializer):
+class JsonSerializer(object):
     name = JSON
 
-    def __init__(self):
-        pass
+    def serialize_header(self, headers):
+        headers = headers or {}
+
+        for k, v in headers.iteritems():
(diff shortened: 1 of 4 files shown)


breerly:
@blampe cc

blampe:
We agreed arg2 for `as=json` should be `map[string]string`, but no one has updated the protocol to reflect that. Why does this keep coming up?

breerly:
I think regardless of the wireformat, we should only accept `map[string]string` from client-code's POV. @blampe we need to have a quick talk w @Raynos and friends to get on the same page. I had thought that email thread a bit back had us there, but we clearly did something wrong.

abhinav:
@Raynos summarized the current state of things here: https://github.com/uber/tchannel/pull/1391#issuecomment-163455903

Raynos:
@abhinav 

This looks great to me. I'm going to fix node :)

tchannel/serializer/json.py
```diff
@@ -23,14 +23,27 @@
...
     name = JSON
 
-    def __init__(self):
-        pass
+    def serialize_header(self, headers):
+        headers = headers or {}
+
+        for k, v in headers.iteritems():
+            if not (isinstance(k, basestring) and isinstance(v, basestring)):
+                raise ValueError(
```
breerly: nice, thanks

breerly:
Looks good to me. AFAICT you'll need to rebase off of master.

PR #340, opened 2015-12-10
Add thrift_request_builder back to the documentation

abhinav:
It's deprecated but people still need to be able to look it up. I added deprecation warnings to the docs.

image


abhinav:
@blampe @breerly @junchaowu

breerly:
Fair enough. :shipit:

blampe:
Now if only people would actually read the docs :)

PR #342, opened 2015-12-10
Fix does-not-exist error if VCR destination directory does not exist

abhinav:


abhinav:
@blampe @breerly @junchaowu

breerly:
Straight-forward. :shipit:

PR #343, opened 2015-12-11
Add spanHost field into tcollector submit

junchaowu:
Based on Yuri's request adding the spanHost field.
@abhinav @blampe @breerly 

cc @prashantv  @Raynos

### DIFF
--- a/tchannel/zipkin/formatters.py
+++ b/tchannel/zipkin/formatters.py
@@ -131,7 +131,7 @@ def i64_to_base64(data):
     return base64.b64encode(i64_to_string(data))
 
 
-def thrift_formatter(trace, annotations, isbased64=False):
+def thrift_formatter(trace, annotations, isbased64=False, span_host=None):
     thrift_annotations = []
     binary_annotations = []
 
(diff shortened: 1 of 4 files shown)


blampe:
@yurishkuro fyi

tchannel/zipkin/tracers.py
```diff
@@ -148,10 +153,18 @@ def submit_callback(f):
                     exc_info=f.exc_info()
                 )
 
+        span_host = self.parse_host_port()
```
yurishkuro: what's the reason not to do it in `__init__`?  We already have tchannel by then.

tchannel/zipkin/formatters.py
```diff
@@ -160,7 +167,9 @@ def thrift_formatter(trace, annotations, isbased64=False):
         host=host,
         parentId=i64_to_string(trace.parent_span_id),
         annotations=thrift_annotations,
-        binaryAnnotations=binary_annotations
+        binaryAnnotations=binary_annotations,
+        debug=False,
```
yurishkuro: a different issue - we should be carrying over the debug flag from the parent span. But it's not critical.

yurishkuro:
lgtm except for one possible performance issue, there's no need to recreate span_host every time

tchannel/zipkin/tracers.py
junchaowu: the zipkin trace hook is added usually before tchannel starts listening. Therefore the tchannel host port is not ready at __init__.

PR #344, opened 2015-12-17
Modified the choose method to use peer heap to select peer

junchaowu:
@abhinav @breerly @blampe 
Change to choose method to use peer heap for selection.

### DIFF
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -23,6 +23,11 @@ from __future__ import division
 
 import math
 import six
+from collections import deque
+
+
+class NoMatchError(Exception):
+    pass
 
 
 class HeapOperation(object):
@@ -46,6 +51,15 @@ class HeapOperation(object):
 
     def pop(self):
         """Pop an item from the heap"""
+        raise NotImplementedError()
+
+    def peek(self, i):
+        """Peek at the item at the given position without removing it from the
+        heap.
+
+        :param i:
+            0-indexed position of the iteam in the heap
+        """
         raise NotImplementedError()
 
     def swap(self, i, j):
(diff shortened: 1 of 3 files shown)


tchannel/tornado/dispatch.py
```diff
@@ -253,6 +250,14 @@ def handle_call(self, request, connection):
 
         raise gen.Return(response)
 
+    def get_endpoint(self, name):
```
abhinav: What's up with this change? It's unrelated to the rest of the PR.

tchannel/tornado/peer.py
```diff
@@ -735,14 +735,19 @@ def choose(self, hostport=None, blacklist=None):
...
-            return self.get(host)
+        while n > 0:
+            peer = self.peer_heap.pop_peer()
+            pop_peers.append(peer)
+            if peer.hostport not in blacklist:
+                choose_peer = peer
+                break
+            n -= 1
+
+        for p in reversed(pop_peers):
```
abhinav: Does this order really matter? The peer heap will take care of putting the peer where it belongs based on its score

tchannel/tornado/peer.py
```diff
@@ -735,14 +735,19 @@ def choose(self, hostport=None, blacklist=None):
...
-        peers = list(self._connected_peers(hosts))
-        if peers:
-            return peers[random.randint(0, len(peers)-1)]
-        else:
-            hosts = list(hosts)
-            host = hosts[random.randint(0, len(hosts)-1)]
-            return self.get(host)
+        while n > 0:
+            peer = self.peer_heap.pop_peer()
+            pop_peers.append(peer)
```
abhinav: This pop-then-push when selecting peers seems inefficient. You can implement a function that finds the "smallest" connected peer without modifying the list in any way.

abhinav: I implemented this in #345 for merging into the this branch. You should be able to use it here.

PR #345, opened 2015-12-18
Add heap.smallest to find the smallest peer that matches a predicate.

abhinav:
Using this should make `PeerGroup.choose` from #344 a lot more efficient than the pop-then-push approach.

### DIFF
--- a/tchannel/container/heap.py
+++ b/tchannel/container/heap.py
@@ -23,6 +23,11 @@ from __future__ import division
 
 import math
 import six
+from collections import deque
+
+
+class NoMatchError(Exception):
+    pass
 
 
 class HeapOperation(object):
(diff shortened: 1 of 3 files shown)


abhinav:
@blampe @breerly @junchaowu

junchaowu:
lgtm

tchannel/container/heap.py
```diff
@@ -23,6 +23,11 @@
 
 import math
 import six
+from collections import deque
```
breerly: nice

tchannel/container/heap.py
```diff
@@ -122,3 +136,42 @@ def down(h, parent, n):
 
         h.swap(parent, min_child)
         parent = min_child
+
+
+def smallest(h, p, current=0):
```
breerly: might be easier on client-code if you expand these variables to be full names.

breerly:
:shipit:  (just some nits)

tchannel/container/heap.py
abhinav: ooh i forgot that `current=0` isn't actually used.

regarding naming: I was trying to be consistent with the rest of the file but I'll expand them.

abhinav:
Updated variable names. Will merge after tests pass.

abhinav:
@junchaowu This is merged now. Feel free to use this in your PR.

PR #349, opened 2015-12-27
FIX the infinite trace submit problem

### LINKED ISSUE #346: downstream requests trigger an infinite loop in zipkin
Apply this diff:
```diff
diff --git a/tchannel/zipkin/zipkin_trace.py b/tchannel/zipkin/zipkin_trace.py
index 59d4db3..127e5e0 100644
--- a/tchannel/zipkin/zipkin_trace.py
+++ b/tchannel/zipkin/zipkin_trace.py
@@ -65,12 +65,14 @@ class ZipkinTraceHook(EventHook):
         return id < self._check_point

     def before_send_request(self, request):
+        print 'sending a req with', request.tracing.traceflags, request.tracing.trace_id, request.tracing.parent_span_id
         if not request.tracing.traceflags:
             return

         if not request.tracing.parent_span_id and not self._lucky(
             request.tracing.trace_id
         ):
+            print 'shedding load'
             # disable trace
             request.tracing.traceflags = False
             return
```

Instrument a simple server to trace 1% of requests:

```python
from tornado import gen
from tornado import ioloop

from tchannel import TChannel
from tchannel.zipkin.zipkin_trace import ZipkinTraceHook

tchannel = TChannel('my-cool-service')


@tchannel.json.register
@gen.coroutine
def endpoint(request):
    yield tchannel.json('some-other-service', 'foo')
    raise gen.Return({"hello": "world"})


@gen.coroutine
def main():
    yield tchannel.advertise(['10.29.39.18:21300'])
    zipkin_tracer = ZipkinTraceHook(tchannel, sample_rate=0.01)
    tchannel.hooks.register(zipkin_tracer)
    tchannel.listen()


if __name__ == '__main__':
    main()
    ioloop.IOLoop.current().start()
```

Hit the endpoint *once* and you'll get an infinite stream of these messages:
```
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
sending a req with True 970968216173609874 970968216173609874
sending a req with True 17281932429783574397 17281932429783574397
```
 
This is using tchannel master.

@junchaowu

junchaowu:
this patch fixes #346

@breerly @blampe @abhinav

### DIFF
--- a/tests/integration/trace/test_zipkin_trace.py
+++ b/tests/integration/trace/test_zipkin_trace.py
@@ -39,7 +39,6 @@ from tchannel.zipkin.trace import Trace, _uniq_id
 from tchannel.zipkin.tracers import TChannelZipkinTracer
 from tchannel.zipkin.zipkin_trace import ZipkinTraceHook
 from tests.mock_server import MockServer
-from tornado.concurrent import Future
 
 try:
     from cStringIO import StringIO
@@ -202,16 +201,10 @@ def test_span_host_field():
 
 @pytest.mark.gen_test
 def test_no_infinite_trace_submit():
+    """Zipkin submissions must not trace themselves."""
 
     def submit(request):
-        count[0] += 1
-        # three time (1 time as client, 2 times as server)
-        if count[0] == 3:
-            f.set_result(count[0])
         return TResponse(True)
-
-    count = [0]
-    f = Future()
 
     zipkin_server = TChannel('zipkin')
     zipkin_server.thrift.register(TCollector, handler=submit)
@@ -219,11 +212,11 @@ def test_no_infinite_trace_submit():
 
     class TestTraceHook(EventHook):
         def __init__(self):
-            self.invalid_traceflag = 0
+            self.tracings = []
 
         def before_send_request(self, request):
-            if request.service == 'tcollector' and request.tracing.traceflags:
-                self.invalid_traceflag += 1
+            # if request.service == 'tcollector':
+            self.tracings.append(request.tracing)
 
     server = TChannel('server', known_peers=[zipkin_server.hostport])
     server.hooks.register(ZipkinTraceHook(tchannel=server, sample_rate=1))
@@ -248,6 +241,7 @@ def test_no_infinite_trace_submit():
 
     client = TChannel('client')
     client.thrift.register(TCollector, handler=submit)
+
     yield client.raw(
         service='server',
         endpoint='hello',
@@ -256,6 +250,11 @@ def test_no_infinite_trace_submit():
         trace=True,
     )
 
-    count = yield f
-    assert count == 3
-    assert test_trace_hook.invalid_traceflag == 0
+    # Continue yielding to the IO loop to allow our zipkin traces to be
+    # handled.
+    for _ in xrange(100):
+        yield tornado.gen.moment
+
+    # One trace for 'hello' and then 3 submissions to tcollector (1 time as
+    # client, 2 times as server)
+    assert len(test_trace_hook.tracings) == 4, test_trace_hook.tracings
--- a/tests/integration/trace/test_zipkin_trace.py
+++ b/tests/integration/trace/test_zipkin_trace.py
@@ -57,6 +57,7 @@ def register(tchannel):
             service='handler2',
             hostport=hostport,
             endpoint="endpoint2",
+            trace=True,
         )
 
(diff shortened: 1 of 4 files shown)


blampe:
When I apply the test to master it still passes, so I don't believe that it's actually testing the fix. 

Additionally I don't understand how this actually addresses the problem. When I was poking around with this it seemed to be an issue with how we determine whether a request's tracing came from the edge. That is, sometimes we have tracing info where `parent_span_id == 0` and sometimes we have `parent_span_id == trace_id` but both of these cases represent the same thing (this is the first span in our trace). Below is a diff that I put together that fixed the issue but I wasn't able to come up with a failing test.

```diff
diff --git a/CHANGES.rst b/CHANGES.rst
index ea3dcb7..042e351 100644
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changes by Version
 0.21.1 (unreleased)
 -------------------

-- Nothing changed yet.
+- Fixed a bug in Zipkin instrumentation that would cause CPU spikes due to an
+  infinite loop during downstream requests.


 0.21.0 (2015-12-10)
diff --git a/tchannel/zipkin/trace.py b/tchannel/zipkin/trace.py
index 923c177..a1a1438 100644
--- a/tchannel/zipkin/trace.py
+++ b/tchannel/zipkin/trace.py
@@ -151,3 +151,11 @@ class Trace(object):
         )

         return trace
+
+    def is_from_edge(self):
+        """Whether this request is the originator of a Trace."""
+        return (
+            self.parent_span_id == self.trace_id
+        ) or (
+            not self.parent_span_id
+        )
diff --git a/tchannel/zipkin/zipkin_trace.py b/tchannel/zipkin/zipkin_trace.py
index 59d4db3..f2da8ab 100644
--- a/tchannel/zipkin/zipkin_trace.py
+++ b/tchannel/zipkin/zipkin_trace.py
@@ -68,7 +68,7 @@ class ZipkinTraceHook(EventHook):
         if not request.tracing.traceflags:
             return

-        if not request.tracing.parent_span_id and not self._lucky(
+        if request.tracing.is_from_edge() and not self._lucky(
             request.tracing.trace_id
         ):
             # disable trace
```

tests/integration/trace/test_zipkin_trace.py
```diff
@@ -236,3 +198,64 @@ def test_span_host_field():
...
     assert thrift_obj.spanHost.serviceName == host_span.service_name
+
+
+@pytest.mark.gen_test
+def test_no_infinite_trace_submit():
+
+    def submit(request):
+        count[0] += 1
+        # three time (1 time as client, 2 times as server)
+        if count[0] == 3:
```
blampe: General feedback: the assertion below is about your test's behavior and not the code you're trying to test. If we still had an infinite loop (submitting millions of traces) this wouldn't catch it because it would stop submitting traces after it's seen 3. It's better to let the code submit all the traces it needs to and then assert that only happened 3 times.

PR #353, opened 2016-01-05
don't start the threadloop until we have work to do

blampe:
@junchaowu @breerly @abhinav

### DIFF
--- a/tchannel/sync/client.py
+++ b/tchannel/sync/client.py
@@ -103,7 +103,6 @@ class TChannel(AsyncTChannel):
 
         )
         self._threadloop = threadloop or ThreadLoop()
-        self._threadloop.start()
 
         self.advertise = self._wrap(self.advertise)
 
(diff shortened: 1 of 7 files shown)


tchannel/sync/client.py
```diff
@@ -115,6 +114,9 @@ def _wrap(self, f):
         assert callable(f)
 
         def wrapper(*a, **kw):
+            if not self._threadloop.is_ready():
```
breerly: I'm not quite sure that this makes sense, Threadloop should block until ready.

breerly: Where were we calling `start` before?

breerly: Dope. I see what you did there. Yeah, this will probably work.

breerly:
Lazy thread, I could dig that. :shipit:

abhinav:
:+1:

tchannel/sync/client.py
```diff
@@ -103,7 +103,6 @@ def __init__(
 
         )
         self._threadloop = threadloop or ThreadLoop()
```
yurishkuro: if the caller passes an externally created threadloop, maybe tchannel should not be trying to start it at all.

blampe: Yeah, this also fixes a bug where the caller could have passed a started thread.

PR #354, opened 2016-01-08
Update choose method to use smallest util function

junchaowu:
@abhinav @blampe @breerly

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -28,7 +28,6 @@ import sys
 from collections import deque
 from itertools import takewhile, dropwhile
 
-from tchannel.container.heap import smallest, NoMatchError
 from tornado import gen
 from tornado.iostream import StreamClosedError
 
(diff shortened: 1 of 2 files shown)


tchannel/tornado/peer.py
```diff
@@ -27,6 +27,8 @@
 
 from collections import deque
 from itertools import takewhile, dropwhile
+
+from tchannel.container.heap import smallest, NoMatchError
```
abhinav: import order?

tchannel/tornado/peer.py
```diff
@@ -735,19 +737,11 @@ def choose(self, hostport=None, blacklist=None):
...
-
-        return choose_peer
+        try:
+            return self.peer_heap.peek(
+                smallest(
+                    self.peer_heap, (lambda p: p.hostport not in blacklist),
+                ),
+            )
+        except NoMatchError:
+            return None
```
abhinav: Wouldn't you rather stick this onto the `PeerHeap` class?

abhinav: To clarify: Up until now, this class depended on only 'self.peer_heap` and its operations. Putting this logic here skips a level from this class' point of view. You're accessing `self.peer_heap` implementation details. You should add a `smallest` method to `PeerHeap` and call that from here.

junchaowu: good idea. I will move it to peerheap class.

abhinav:
k

PR #355, opened 2016-01-12
Heap selection branch merged into mash develop branch.

junchaowu:
The perf benchmark from mash branch:
screen shot 2016-01-11 at 9 05 56 pm

The perf bench mark from heap branch:
screen shot 2016-01-11 at 9 08 45 pm
@abhinav @breerly @blampe


benchmarks/test_peer_selection.py
```diff
@@ -22,81 +22,78 @@
     absolute_import, unicode_literals, print_function, division
 )
 
-import mock
-import random
+from tornado import gen, ioloop
```
abhinav: This looks like a completely different benchmark. Why are you replacing the existing benchmark with this instead of creating a new one?

tchannel/peer_heap.py
```diff
@@ -0,0 +1,157 @@
...
+        Primary comparator will be the score of each peer. If the ``score`` is
+        same then compare the ``order``. The ``order`` attribute of the peer
+        tracks the heap push order of the peer. This help solve the imbalance
+        problem caused by randomization when deal with same score situation.
+
+        :param i: ith peer
+        :param j: jth peer
+        :return: True or False
+        """
+        if self.peers[i].score == self.peers[j].score:
```
abhinav: I mentioned this in the original PR too: Calling this score is really confusing because that implies that high score means the peer is more desirable, while the behavior we are actually using is that low score means that the peer is more desirable. I've already been tripped by that in the previous code reviews. I highly recommend renaming `score` to something that implies that lower is better. You mentioned that `pending`/`ongoing` (as in the number of ongoing requests) didn't work here because we'll eventually want more complex criteria for measuring the score. How about `rank`? Or we could still call it `ongoing` or `pending` and rename it if and when we add more complex ranking criteria?

benchmarks/test_peer_selection.py
blampe: Are the perf numbers in the summary master vs. this branch running old vs. new benchmark or are they both running this benchmark? Need apples-to-apples comparisons.

junchaowu: yes they both are using this benchmark. I compared the mash branch and the
heap branch. The different between these two branch is just the peer heap
change. In order to make it exact apple to apple comparison , we also
change the peer choose in mash branch to choose the least pending request's
peer so keep them using the same branch.

junchaowu: Strategy*

tchannel/peer_heap.py
junchaowu: change the score to rank.

benchmarks/test_peer_selection.py
junchaowu: The original benchmark doesn't work for the new case, since the original one just keep calling peer's choose method. In peer heap's branch , the choose method will simply fetch the top of the heap which doesn't make sense. It is not measuring the whole peer heap logic, it just measures the heap's peek call.

abhinav: Yes, I understand that but this is measuring something different.

I want to leave the benchmark that measures the cost of calling `choose()` separate from your new benchmark. The cost of choose used to be `O(n)`, and now it is `O(1)`. Maybe in the future we change it to `O(log n)` because it allows us to optimize something else elsewhere in the system. I want this benchmark, that measures the cost of `choose()` to stay in the repo so that we can use it if necessary.

The new benchmark that makes end to end requests is a separate benchmark that goes in a different file.

tchannel/tornado/peer.py
```diff
@@ -578,7 +629,9 @@ def get(self, hostport):
         """
         assert hostport, "hostport is required"
         if hostport not in self._peers:
-            self._peers[hostport] = self.peer_class(self.tchannel, hostport)
+            self.add(hostport)
+            return self._peers[hostport]
+
         return self._peers[hostport]
```
abhinav: drop the return from the if statement. you're already returning the same result on the next line.

tchannel/tornado/peer.py
```diff
@@ -608,13 +664,28 @@ def add(self, peer):
 
         if isinstance(peer, basestring):
             # Assume strings are host-ports
-            peer = self.peer_class(self.tchannel, peer)
+            peer = self.peer_class(
+                tchannel=self.tchannel,
+                hostport=peer,
+                on_conn_change=self.update_heap,
+            )
+            peer.rank = self.rank_calculator.get_rank(peer)
```
abhinav: You probably need to set the rank on peers that you're not constructing on demanding but adding manually as well, don't you?

tchannel/tornado/peer.py
```diff
@@ -608,13 +664,28 @@ def add(self, peer):
...
 
         assert peer.hostport not in self._peers, (
             "%s already has a peer" % peer.hostport
         )
 
         self._peers[peer.hostport] = peer
+        self.peer_heap.push_peer(peer)
+
+    def update_heap(self, peer):
```
abhinav: Make this protected? (Start with `_`.)

tchannel/peer_strategy.py
```diff
@@ -0,0 +1,64 @@
...
+
+class RankCalculator(object):
+    """rankCalculator provides interface to calculate the rank of a peer."""
+    def get_rank(self, peer):
+        raise NotImplementedError()
+
+
+class ZerorankCalculator(RankCalculator):
+    def get_rank(self, peer):
+        return 0
```
abhinav: Is this safe to delete? Doesn't appear to be used anywhere.

tchannel/peer_strategy.py
```diff
@@ -0,0 +1,66 @@
...
+    """RankCalculator calculates the rank of a peer."""
+
+    def get_rank(self, peer):
+        raise NotImplementedError()
+
+
+class ZerorankCalculator(RankCalculator):
+
+    def get_rank(self, peer):
+        return 0
```
abhinav: Is this safe to delete? Doesn't appear to be used anywhere.

junchaowu: Though it is not used now, it will be helpful to keep it. The zero calculator is used for debug purpose by giving us same rand for all peers.

abhinav:
LGTM

tchannel/peer_strategy.py
abhinav: Discussed offline. Deleting.

PR #358, opened 2016-01-13
Add to/from_dict for req/res TransportHeaders

### LINKED ISSUE #350: TransportHeaders.to_dict()
We need an easy way for users to convert transport headers to/from the key-value dict that will actually be sent over the wire. A use case for this is if you're implementing a catch-all endpoint that needs to serialize the contents of the request for some reason (to send them to another server, for example), it also needs to record the transport headers. 

Side: This will be needed for VCR anyway when we actually remove the deprecated API.

abhinav:
Resolves #350.

### DIFF
--- a/tchannel/response.py
+++ b/tchannel/response.py
@@ -23,6 +23,7 @@ from __future__ import (
 )
 
 from . import schemes
+from . import transport as t
 from .status import OK
 
 __all__ = ['Response']
@@ -74,16 +75,30 @@ class TransportHeaders(object):
         'scheme',
     )
 
-    def __init__(self,
-                 failure_domain=None,
-                 scheme=None,
-                 **kwargs):
-
+    def __init__(self, failure_domain=None, scheme=None):
         if scheme is None:
             scheme = schemes.RAW
 
         self.failure_domain = failure_domain
         self.scheme = scheme
+
+    @classmethod
+    def from_dict(cls, data):
+        return cls(
+            failure_domain=data.get(t.FAILURE_DOMAIN),
+            scheme=data.get(t.SCHEME),
+        )
+
+    def to_dict(self):
+        m = {}
+
+        if self.failure_domain is not None:
+            m[t.FAILURE_DOMAIN] = self.failure_domain
+
+        if self.scheme is not None:
+            m[t.SCHEME] = self.scheme
+
+        return m
 
 
 def response_from_mixed(mixed):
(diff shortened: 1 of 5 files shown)


abhinav:
@blampe @breerly @junchaowu

tchannel/response.py
```diff
@@ -74,17 +75,31 @@ class TransportHeaders(object):
         'scheme',
     )
 
-    def __init__(self,
-                 failure_domain=None,
-                 scheme=None,
-                 **kwargs):
```
breerly: Removing this would be breaking right?

junchaowu: that's breaking.

abhinav: No, it's not breaking. This was in place because `to_kwargs` would send in a dict with `None` values for headers that were not valid transport headers for response but were valid for request.

Since `from_dict` is now responsible for instantiating `TransportHeaders` you will never get unexpected kwargs there.

breerly: Good point. We're cool.

tchannel/tchannel.py
```diff
@@ -207,12 +207,10 @@ def call(
         # unwrap response
         body = yield response.get_body()
         headers = yield response.get_header()
-        t = transport.to_kwargs(response.headers)
-        t = TransportHeaders(**t)
         result = Response(
             body=body,
             headers=headers,
-            transport=t
+            transport=TransportHeaders.from_dict(response.headers),
```
breerly: nit - would rather not see a function call while instantiating an object

breerly:
LGTM. :shipit:

tchannel/response.py
junchaowu: This a risky part in python. We can't prevent others using ``TransportHeaders`` directly. Therefore though we can make sure we pass valid arguments, we can't guarantee the behavior that whoever try to instantiate the ``TransportHeaders`` directly.

abhinav: Yeah but it's safe to assume that people aren't messing with this directly because:

1. It's not a documented public API
2. We don't provide a way to inject one that you construct yourself into a request or response.

We can still add it to the changelog just in case, and maybe release it in a minor version bump.

blampe:
+1

tchannel/tchannel.py
abhinav: true

PR #359, opened 2016-01-19
Remove the exception in the register method in sync tchannel.

junchaowu:
@blampe @abhinav 
consider a more user friendly behavior , replace the exception in sync tchannel register to log message.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.21.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Attempting to register endpoints against a synchronous TChannel is now a no-op instead of a crash.
 
 
 0.21.2 (2016-01-05)
--- a/tchannel/sync/client.py
+++ b/tchannel/sync/client.py
@@ -20,9 +20,14 @@
 
 from __future__ import absolute_import
 
+import logging
+
 from threadloop import ThreadLoop
 
 from tchannel import TChannel as AsyncTChannel
+
+
+log = logging.getLogger('tchannel')
 
 
 class TChannel(AsyncTChannel):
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Changes by Version
 0.21.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Make ``tchannel.sync.client.register`` no ops.
```
abhinav: > Attempting to register endpoints against a synchronous TChannel is now a no-op instead of a crash.

tchannel/sync/client.py
```diff
@@ -104,27 +109,49 @@ def __init__(
         )
         self._threadloop = threadloop or ThreadLoop()
 
+        self.__init__ = self._wrap(self.__init__)
```
abhinav: I don't think this does what you want it do. What are you trying to do here?

tchannel/sync/client.py
```diff
@@ -104,27 +109,48 @@ def __init__(
...
+
+    def register(self, *args, **kwargs):
+        return _register(*args, **kwargs)
+
+
+def _register(*args, **kwargs):
+    log.warning("Registration not yet supported for sync tchannel.")
+
+    def decorator(fn):
+        pass
```
abhinav: change this to `return fn`. Otherwise, the decorated function gets replaced with `None` for the caller.

tchannel/sync/client.py
junchaowu: right. no need.

tchannel/sync/client.py
```diff
@@ -104,27 +109,48 @@ def __init__(
...
+
+    return decorator
+
+
+def _sync_to_async(threadloop, func, *args, **kwargs):
+    if not threadloop.is_ready():
+        threadloop.start()
+    return threadloop.submit(lambda: func(*args, **kwargs))
```
abhinav: You can change this to,

    threadloop.submit(func, *args, **kwargs)

tchannel/sync/client.py
```diff
@@ -104,27 +109,48 @@ def __init__(
...
+            return _sync_to_async(self._threadloop, f, *a, **kw)
 
         return wrapper
 
-    def register(self, *a, **kw):
-        raise NotImplementedError(
-            "Registration not yet supported for sync clients",
-        )
+
+class TChannelSyncWrapper(object):
```
abhinav: Call this `_SyncScheme` or something similar.

tchannel/sync/client.py
```diff
@@ -104,27 +109,48 @@ def __init__(
...
+def _register(*args, **kwargs):
+    log.warning("Registration not yet supported for sync tchannel.")
+
+    def decorator(fn):
+        pass
+
+    return decorator
+
+
+def _sync_to_async(threadloop, func, *args, **kwargs):
```
abhinav: Well, technically it's async to sync. I would just call it `_submit` or something. Although, I'm not sure you even need this since you can just do the check and submit manually in two places you're calling this.

tests/sync/test_client.py
```diff
@@ -115,3 +114,61 @@ def test_should_discover_ip():
...
+    client = TChannel('test-client')
+
+    @client.json.register
+    def hello(request):
+        return {}
+
+    assert not client._dep_tchannel._handler.handlers.get('hello', None)
+
+    client.json.register(hello)
+    assert not client._dep_tchannel._handler.handlers.get('hello', None)
```
abhinav: This seems like a janky way to test this. With that number of `_` and `.`, there has to be a better way.

Try creating a parameterized fixture (see http://pytest.org/latest/example/parametrize.html) that returns a function that takes a tchannel and registers endpoints on it. So, parameterize the fixture over whether you want raw/json/thrift endpoints (and perhaps a "direct" one which just uses tchannel.register) and return the appropriate function.

Then in the test, call that function on one sync tchannel and one async tchannel. Ensure that you're able to make requests to the async tchannel and that requests to the sync tchannel fail with unknown method or whatever we have there.

junchaowu:
squash all the commits

abhinav:
@junchaowu Why does your squashed version contain commits from me. I think you messed up the history of your branch somehow.

tests/sync/test_client.py
```diff
@@ -115,3 +114,96 @@ def test_should_discover_ip():
...
+    async_client = AsyncTchannel('s')
+    with pytest.raises(BadRequestError):
+        yield async_client.json(
+            service='test-client',
+            endpoint='hello',
+            hostport=sync_client.hostport
+        )
+
+    sync_client.json.register(hello)
+    with pytest.raises(BadRequestError):
```
abhinav: Can you not duplicate these tests like this? It's not very difficult to use fixtures for this. Declare these fixtures:

- scheme: Just a string value. One of json, thrift, raw.
- register_endpoints(scheme): returns a function that takes a tchannel and registers endpoints against it. could optionally be parameterized over whether you use a decorator or a manual `.register` call.
- call(scheme): Returns a function that makes a request to endpoint defined in register_endpoints for the specific scheme.

Then write two separate tests, both using register_endpoints and call: one with a sync tchannel, one with an async tchannel. This allows you to test the 3 * 2 * 2 approaches with two tests rather than 12.

blampe:
@junchaowu bad rebase?

junchaowu:
solved the rebase issue.

abhinav:
:+1:

PR #365, opened 2016-02-12
Leave short-lived markers behind when requests time out locally.

abhinav:
These markers ensure that if we receive a message for those requests shortly
after the local timeout, we know that we can safely ignore those requests.


abhinav:
CC @blampe @breerly @junchaowu @jcorbin

jcorbin:
This looks like valid python to me, and similar enough to https://github.com/uber/tchannel-node/blob/master/operations.js#L98 to be what you want.

blampe:
Cool.

tchannel/tornado/tombstone.py
```diff
@@ -0,0 +1,108 @@
...
+
+A tombstone is a marker for a request that timed out. Often, shortly after a
+request times out, we receive a message for it anyway. This is because the
+server's timeout is often skewed. So if we immediately forget about the
+request on timeout, we'll see an "unconusmed message" error when we receive
+these zombie messages. To avoid this, we leave a tombstone behind for the
+request that timed out so that we know it is safe to ignore the zombie
+messages.
+
+Tombstones are destroyed automatically after a fixed duration.
```
breerly: Nice.

tests/tornado/test_connection.py
```diff
@@ -81,3 +83,36 @@ def test_close_callback_is_called():
...
+            timeout=0.05, hostport=server.hostport,
+            # The server will take 70 milliseconds but we allow at most 50.
+        )
+
+    # Wait for the server to send the late response and make sure it doesn't
+    # log a warning.
+    with mock.patch.object(connection.log, 'warn') as mock_warn:  # :(
+        yield gen.sleep(0.03)
+
+    assert mock_warn.call_count == 0
```
breerly: this seems fragile...

breerly:
I'm ok with :shipit: here.

nit: testing the absence of behavior is usually a smell, although i dont have a better immediate alternative at this moment.

junchaowu:
:ship:

tests/tornado/test_connection.py
abhinav: yeah.. :disappointed:

PR #368, opened 2016-03-01
Serialize writes

blampe:
Say a connection is writing two fragmented responses concurrently. If response A writes a fragment before response B finished writing its fragment, then B's write future never completes and the client will time out waiting for B.

```
If no callback is given, this method returns a Future that resolves (with a result of None) when the write has been completed. If write is called again before that Future has resolved, the previous future will be orphaned and will never resolve.
```

There's a performance hit here by using a coroutine in the write path, but at least this reduces us to one writer coroutine instead of every request handling its own writing. I don't have a good test for this but I was able to confirm it fixed the issue on @junchaowu's `debug` branch.

@abhinav @junchaowu


tchannel/tornado/connection.py
```diff
@@ -358,8 +365,15 @@ def _write(self, message):
...
         body = frame.frame_rw.write(f, BytesIO()).getvalue()
+        return self._write_queue.put(body)
 
-        return self.connection.write(body)
+    @tornado.gen.coroutine
+    def _drain_write_queue(self):
+        """Ensures ordering of our writes."""
+
+        while not self.closed:
+            body = yield self._write_queue.get()
```
junchaowu: problem with this `get` is it will have timeout. It blocks until timeout. So we need to catch the timeout .

blampe: It has a timeout of None so blocks indefinitely until something is available to write. http://www.tornadoweb.org/en/stable/queues.html#tornado.queues.Queue.get

tchannel/tornado/connection.py
```diff
@@ -358,8 +365,15 @@ def _write(self, message):
...
+            yield self.connection.write(body)
```
abhinav: this will stop writing messages indefinitely when one of them fails

tchannel/tornado/connection.py
```diff
@@ -138,6 +137,9 @@ def __init__(self, connection, tchannel=None, direction=None):
         self.tchannel = tchannel
         self._close_cb = None
 
+        self._write_queue = tornado.queues.Queue()
```
abhinav: Consider adding a maxsize to create back pressure on the callers and ensure the queue doesn't grow unreasonably

tchannel/tornado/connection.py
```diff
@@ -358,8 +365,15 @@ def _write(self, message):
             payload=payload
         )
         body = frame.frame_rw.write(f, BytesIO()).getvalue()
+        return self._write_queue.put(body)
 
-        return self.connection.write(body)
+    @tornado.gen.coroutine
```
abhinav: You can drop the `gen.coroutine` by making this future-callback based.

```python
def _drain_write_queue(self):
    io_loop = IOLoop.current()

    def _on_write(future):
        io_loop.spawn_callback(_write_next)
        if future.exception():    
            log(future.exception())

    def _on_message(future):
        if future.exception():
            io_loop.spawn_callback(_write_next)
            log(future.exception())
            return
        message = future.result()
        io_loop.add_future(self.connection.write(message), _on_write)

    def _write_next():
        if self.closed:
            return
        io_loop.add_future(self._write_queue.get(), _on_message)

    io_loop.spawn_callback(_write_next)
```

Note that the use of `spawn_callback` makes it drop any stack context knowledge during the write. That should reduce overhead. If we want to keep that information, we'll have to use `add_callback`.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
         # while.
         future.set_exception(errors.TimeoutError())
         self._request_tombstones.add(req_id, req_ttl)
+
+
+class Reader(object):
+
+    def __init__(self, io_stream, capacity=64):
```
abhinav: Nit: Always use None for default args.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
         # while.
         future.set_exception(errors.TimeoutError())
         self._request_tombstones.add(req_id, req_ttl)
+
+
+class Reader(object):
```
abhinav: Some docstrings for the different methods on this would be nice.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+        self.io_stream = io_stream
+
+    def fill(self):
+        self.filling = True
+
+        io_loop = IOLoop.current()
+
+        def keep_reading(f):
+            if f.exception():
+                return log(f.exception())
```
abhinav: You want to spawn a callback even in case of an exception, don't you? Or do you want it to stop filling once an error happens. Also, you should set filling to False if you stop.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+
+            io_loop.add_future(
+                read_body_future,
+                lambda future: on_body(future, size)
+            )
+
+            return read_body_future
+
+        def on_error(future):
+            log.info("Failed to read data: %s", future.exception())
```
abhinav: You need to also transfer the exception to the message_future in case of error otherwise it'll never resolve.

    message_future.set_exc_info(future.exc_info())

This will preserve the stack trace too.

abhinav:
An issue with the future callback style is that non-IO exceptions must be caught by you. `@coroutine` usually does a catch-all. I would suggest using something like this: https://gist.github.com/abhinav/601704fcb7761f824006

It will allow you to have uncaught exceptions transferred to the future your final future.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+            message.id = f.header.message_id
+            message_future.set_result(message)
+
+        def on_read_size(read_size_future):
+            if read_size_future.exception():
+                return on_error(read_size_future)
+
+            size_bytes = read_size_future.result()
+            size = frame.frame_rw.size_rw.read(BytesIO(size_bytes))
+            read_body_future = self.io_stream.read_bytes(size - size_width)
```
abhinav: Note that if any of this fails with an exception, `message_future` just won't get filled. See my other comment in the PR comments for how to solve this better.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+                read_body_future,
+                lambda future: on_body(future, size)
+            )
+
+            return read_body_future
+
+        def on_error(future):
+            log.info("Failed to read data: %s", future.exception())
+
+        size_width = frame.frame_rw.size_rw.width()
```
abhinav: This is constant. Don't need to recalculate it every time.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+        :returns:
+            A Future that produces a Context object containing the next
+            message off the wire.
+        """
+        if self.filling is False:
+            self.fill()
+
+        return self.queue.get()
+
+    def _dequeue(self):
```
abhinav: This mostly has nothing to do with queues. It can be moved into a separate function outside the class.

tchannel/tornado/connection.py
```diff
@@ -701,3 +633,162 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+
+    def drain(self):
+        self.draining = True
+
+        io_loop = IOLoop.current()
+
+        def on_write(f, done):
+            if f.exception():
+                log.error(f.exception())
+                done.set_exception(f.exception())
```
abhinav: Prefer `set_exc_info` over `set_exception` to preserve stack traces in Python 2.

abhinav: Prefer `set_exc_info` over `set_exception` to preserve stack traces in Python 2.

abhinav:
Created #371 with the `@fail_to` thing and tests for it to be merged into this branch.

tchannel/tornado/connection.py
blampe: I'm pretty sure this is preserving existing behavior, probably because the connection is in a bad state if this happens. I'll set `filling = False` which will have the effect of re-starting the reader as needed.

tchannel/tornado/connection.py
```diff
@@ -701,3 +634,176 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+        self.filling = True
+
+        io_loop = IOLoop.current()
+
+        def keep_reading(f):
+            if f.exception():
+                self.filling = False
+                return log(f.exception())
+            # connect these two in the case when put blocks
+            self.queue.put(f.result())
```
abhinav: If you don't wait on the future returned by `queue.put` before calling `fill` again, specifying a capacity when you construct the queue is useless. This won't slow down reading, it'll just keep queueing up futures in the IOLoop that will eventually `put` into the queue. You want to do something like this:

```python
self.queue.put(f.result()).add_done_callback(lambda f: io_loop.spawn_callback(self.fill))
# It's safe to use add_done_callback here because spawn_callback already delays the call
# until the next iteration of the IOLoop.
```

(Or also handle the exception for `queue.put`, although I'm not aware of any recoverable exceptions raised by `queue.put`.)

tchannel/tornado/connection.py
```diff
@@ -701,3 +634,176 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+            io_loop.spawn_callback(self.fill)
+
+        io_loop.add_future(read_message(self.io_stream), keep_reading)
+
+    def get(self):
+        """Receive the next message off the wire.
+
+        :returns:
+            A Future that produces a Context object containing the next
```
abhinav: This is outdated. We don't have Context objects in this class anymore (I think).

tchannel/tornado/connection.py
```diff
@@ -701,3 +634,176 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+            message off the wire.
+        """
+        if self.filling is False:
```
abhinav: nit:

    if not self.filling:

tchannel/tornado/connection.py
```diff
@@ -701,3 +634,176 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+
+    def put(self, message):
+        """Enqueues the given message for writing to the wire.
+
+        The message must be small enough to fit in a single frame.
+        """
+        if self.draining is False:
+            self.drain()
+
+        return self._enqueue(message)
```
abhinav: It doesn't look like this future is being waited-on by the callers. Defeats the purpose of having a maxsize since you'll end up queueing a lot of futures on the IOLoop and the order there is undefined (I think). If we can't ensure that callers all the way up the call chain are waiting on the `put`, might as well drop the `capacity`.

tchannel/tornado/connection.py
```diff
@@ -701,3 +634,176 @@ def _request_timed_out(self, req_id, req_ttl, future):
...
+            ),
+            payload=payload
+        )
+        body = frame.frame_rw.write(f, BytesIO()).getvalue()
+
+        done_writing_future = tornado.gen.Future()
+
+        def on_queue_error(f):
+            if f.exception():
+                done_writing_future.set_exc_info(f.exc_info())
```
abhinav: You need to `set_result(None)` on `done_writing_future` or it will never resolve in the success case.

tests/conftest.py
```diff
@@ -47,7 +47,7 @@ def connection():
 
 
 @pytest.yield_fixture
-def mock_server():
+def mock_server(io_loop):
```
abhinav: heh that must have been fun to debug

tests/tornado/test_connection.py
```diff
@@ -39,33 +38,33 @@ def dummy_headers():
...
-    @pytest.fixture(autouse=True)
-    def make_server_client(self, tornado_pair):
-        self.server, self.client = tornado_pair
+@pytest.mark.gen_test
+def test_handshake(tornado_pair):
+    """Verify we handshake in an async manner."""
+    server, client = tornado_pair
+    headers = dummy_headers()
+
+    print 'IOLOOP from test is', tornado.ioloop.IOLoop.current()
```
abhinav: lol

tchannel/tornado/connection.py
blampe: If you `set_result` here then the caller will only wait as long as it takes to insert into the queue.

In the success case `done_writing_future` gets enqueued and resolves in `on_write` once we've actually written bytes on the wire.

tchannel/tornado/connection.py
blampe: It is getting waited on. `connection.write` returns a future that resolves once the last fragment has been written, and that's waited on in `_stream`.

tchannel/tornado/connection.py
abhinav: Oh right. My bad.

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ Changes by Version
 
 - Tornado 4.2 was listed as a requirement but this was corrected to be 4.3
   which introduced the locks module.
+- Fixed in issue where clients could incorrectly time out when reading large
+  response bodies.
```
abhinav: Does this accurately reflect this change? Probably worth mentioning that writes will be serialized now or whatever bug that's fixing?

abhinav:
LGTM :shipit:

PR #370, opened 2016-03-06
Add a roundtrip benchmark

blampe:
I want something that gives us some measurable data on performance. I'm going to be killing some coroutines and I want to be able to confirm that it's helping.

@abhinav @breerly

### DIFF
--- a/benchmarks/test_roundtrip.py
+++ b/benchmarks/test_roundtrip.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


benchmarks/test_roundtrip.py
```diff
@@ -0,0 +1,57 @@
...
+    loop = ioloop.IOLoop.current()
+
+    server = TChannel('benchmark-server')
+    server.listen()
+
+    clients = [TChannel('benchmark-client') for _ in range(10)]
+
+    service = thrift.load(
+        path='examples/guide/keyvalue/service.thrift',
+        service='benchmark-server',
```
abhinav: Safe to load at import time.

benchmarks/test_roundtrip.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    service.hostport = server.hostport
```
abhinav: Provide hostport at load time or at call site.

abhinav:
LGTM besides the hostport thing

PR #377, opened 2016-03-11
connection.Writer.write() should re-raise exceptions

abhinav:
Other changes:

- Fix deprecation warning from Hypothesis during tests
- A test was using the wrong Response class for its mocks
- Removed deprecation warnings because of usage of `BaseException.message`
- Include stack traces while logging in more places

@blampe @breerly @junchaowu

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changes by Version
 0.21.9 (unreleased)
 -------------------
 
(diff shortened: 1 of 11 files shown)


blampe:
changelog then land

abhinav:
Updated

junchaowu:
lgtm

PR #378, opened 2016-03-11
Fix unconsumed StreamClosedError logs

abhinav:
This fixes two places which were causing unconsumed Future exceptions:

1.  `stream_request`'s future simply wasn't consumed
2. During tests, the IOLoop was being stopped before the callback given to `read_message(...).add_future(...)` could be called. I think this same issue would affect people using `IOLoop.run_sync`. The fix is simply to use `add_done_callback` in this case.

### DIFF
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -663,7 +663,7 @@ class Reader(object):
                 lambda f: io_loop.spawn_callback(self.fill),
             )
 
-        io_loop.add_future(read_message(self.io_stream), keep_reading)
+        read_message(self.io_stream).add_done_callback(keep_reading)
 
     def get(self):
         """Receive the next message off the wire.
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -555,13 +555,12 @@ class StreamConnection(TornadoConnection):
         """send the given request and response is not required"""
         request.close_argstreams()
 
+        def on_done(future):
+            future.exception()  # < to avoid "unconsumed exception" logs
+            request.close_argstreams(force=True)
+
         stream_future = self._stream(request, self.request_message_factory)
-
-        IOLoop.current().add_future(
-            stream_future,
-            lambda f: request.close_argstreams(force=True)
-        )
-
+        stream_future.add_done_callback(on_done)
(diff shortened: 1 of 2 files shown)


abhinav:
@blampe @breerly @junchaowu

tchannel/tornado/connection.py
```diff
@@ -555,13 +555,12 @@ def stream_request(self, request):
...
         request.close_argstreams()
 
-        stream_future = self._stream(request, self.request_message_factory)
-
-        IOLoop.current().add_future(
-            stream_future,
-            lambda f: request.close_argstreams(force=True)
-        )
+        def on_done(future):
+            future.exception()  # < to avoid "unconsumed exception" logs
```
junchaowu: so just silently skip?

abhinav: Yeah, same thing we were doing before. I'm reluctant to add new possibly noisy log entries in this part of the system.

tchannel/tornado/connection.py
```diff
@@ -664,7 +663,7 @@ def keep_reading(f):
                 lambda f: io_loop.spawn_callback(self.fill),
             )
 
-        io_loop.add_future(read_message(self.io_stream), keep_reading)
```
blampe: Why `add_done_callback` instead of `add_future`?

abhinav: Explained in the description but basically the `add_future` is the reason that the `read_message` future does not get consumed.

    io_loop.add_future(f, cb) == io_loop.add_callback(lambda: f.add_done_callback(cb))

If the IOLoop is stopped before the next iteration, it never calls `add_done_callback` and `f` stays unconsumed. 

Why the IOLoop is getting stopped before then, I'm not entirely sure, but I believe it might be due to timeouts causing a higher level Future to finish early.

blampe: Hm... ok but I'm skeptical.

blampe:
:ship:

tchannel/tornado/connection.py
abhinav: Agreed. I'm okay with this change mostly because it makes those log messages go away and it won't behave any differently in other cases.

Issue #319, opened 2015-12-02
TChannel.advertise should be safe to call multiple times


abhinav:
Currently, advertise for the sync and async clients do not keep track whether they're already advertising. This means that calling advertise multiple times is unsafe because that'll just start up multiple advertise loops. We should keep track of advertisement status so that it's safe to all advertise multiple times.

breerly:
Not totally sure if we need this. If you really want idempotence, couldn't you just:

```python
from tchannel.sync.singleton import TChannel

if not TChannel.prepared:
    TChannel.prepare(...)
    TChannel.get_instance().advertise(...)
```

and then in other parts of the application you can just:

```python
from tchannel.singleton import TChannel

tchannel = TChannel.get_instance()
```

breerly:
Is there value to making advertise idempotent outside of the singleton facility?

abhinav:
Resolved by #319

PR #379, opened 2016-03-12
Make TChannel.advertise() idempotent and thread safe.

abhinav:
Resloves #319.

### DIFF
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -121,6 +121,11 @@ class TChannel(object):
         # register default health endpoint
         self.thrift.register(Meta)(health)
 
+        # advertise_response is the Future containing the response of calling
+        # advertise().
+        self._advertise_response = None
+        self._advertise_lock = Lock()
+
     def is_listening(self):
         return self._dep_tchannel.is_listening()
 
@@ -260,7 +265,6 @@ class TChannel(object):
         else:
             return decorator(handler)
 
-    @gen.coroutine
     def advertise(self, routers=None, name=None, timeout=None,
                   router_file=None):
         """Advertise with Hyperbahn.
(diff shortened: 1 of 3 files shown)


abhinav:
@blampe @breerly @junchaowu

blampe:
Awesome :sheep:

tchannel/tchannel.py
```diff
@@ -315,14 +322,36 @@ def advertise(self, routers=None, name=None, timeout=None,
...
+                    self._advertise_response = None
+
+        is_new = False
+        with self._advertise_lock:
+            if self._advertise_response is None:
+                self._advertise_response = _advertise()
+                is_new = True
+            future = self._advertise_response
+
+        if is_new:
```
junchaowu: can you move add_done_callback into the 352 line, so you remove is_new.

abhinav: I don't want to call the `add_done_callback` while I have the lock acquired. `add_done_callback` calls the function right away if the future has already resolved. If that happens, the callback will also try to acquire the lock and cause a deadlock since the lock is not re-entrant.

abhinav: Added a comment explaining reasoning in the code.

junchaowu: Even the future is resolved, it will still wait your current method to be finished.

abhinav: No it won't. That's if you use `add_future`. If you use `add_done_callback`, it will call the callback right away as part of the `add_done_callback` call. See http://tornado.readthedocs.org/en/latest/_modules/tornado/concurrent.html#Future.add_done_callback.

breerly:
Thanks for getting rid of `is_new` - this is much cleaner!

breerly:
:shipit:

junchaowu:
lgtm

PR #380, opened 2016-03-15
connection: read_message should propagate StreamClosedError

abhinav:
If the IOStream has already been closed by the time read_message is called, it
will fail with a synchronous StreamClosedError which read_message needs to
catch.

### DIFF
--- a/tests/tornado/test_connection.py
+++ b/tests/tornado/test_connection.py
@@ -164,3 +164,23 @@ def test_writer_write_error():
     writer.io_stream.close()
     with pytest.raises(StreamClosedError):
         yield writer.put(messages.PingResponseMessage())
+
+
+@pytest.mark.gen_test
+def test_reader_read_error():
+    server, client = socket.socketpair()
+    reader = connection.Reader(IOStream(server))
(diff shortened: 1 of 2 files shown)


tests/tornado/test_connection.py
```diff
@@ -164,3 +164,23 @@ def test_writer_write_error():
...
+    assert isinstance(ping, messages.PingRequestMessage)
+
+    reader.io_stream.close()
+    future = reader.get()
+    with pytest.raises(gen.TimeoutError):
+        yield gen.with_timeout(timedelta(milliseconds=100), future)
+
+    # TODO(abg): Need to fix Reader.get to actually raise StreamClosedError
+    # rather than freezing forever. Then this can be changed to assert that a
+    # StreamClosedError is raised.
```
abhinav: Note that this PR is not breaking anything here. This is the **current** behavior with or without this PR. I have a fix incoming for this in a different PR since it's slightly more involved.

abhinav:
Also note that this doesn't have a changelog because it doesn't change behavior from the user's point of view yet.

blampe:
:ship: very nice

PR #381, opened 2016-03-15
Make Reader.get() fail properly in case of errors

abhinav:
When connection.Reader is reading a message, any failures result in the fill-loop stopping. So the `get()` that started the fill loop will just never resolve and report a hard-to-debug `TimeoutError`. We need a way to send back the actual error to the `get()` call.

To do this, I swapped out the async Queue implementation with a custom one that behaves and is implemented very similar to Tornado's version, plus it provides a `terminate(e)` method which fails all outstanding `get` and `put` requests with the given exception. It doesn't provide `join` or track the number of unfinished tasks but we aren't using those features anyway. For reference, here's Tornado's implementation: http://www.tornadoweb.org/en/latest/_modules/tornado/queues.html#Queue


junchaowu:
:+1:

tests/tornado/test_connection.py
```diff
@@ -126,8 +125,8 @@ def test_stream_closed_error_on_read(tornado_pair):
     client.close()
 
     with mock.patch.object(connection, 'log') as mock_log:  # :(
-        with pytest.raises(gen.TimeoutError):
-            yield gen.with_timeout(timedelta(milliseconds=100), future)
+        with pytest.raises(StreamClosedError):
```
blampe: So much better.

blampe:
Is the queue thread-safe? We're currently getting those assertion "getters waiting" errors and I'm wondering what that will look like with this implementation.

:ship: if you're confident this is backwards-compatible and thread safety isn't a concern.

abhinav:
This isn't thread-safe. We can make it thread safe but I was trying to go for a lock-free structure since single-threaded is the default mode of usage for this. There aren't any in-code assertions so instead of the "getters waiting" failure, we'll have those messages processed out of order.

tests/tornado/test_connection.py
breerly: nice

abhinav:
Update: Merged the thread safe simpler implementation of the Queue and rebased on master.

junchaowu:
Yeah give it a try.

abhinav:
Rebased on #386 (which will soon be on master)

breerly:
:ship:

breerly:
Please rebase before merge

PR #382, opened 2016-03-16
Include a cn annotation for all spans

blampe:
@prashantv noticed this was missing from python spans.

@yurishkuro @junchaowu @abhinav

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.21.10 (unreleased)
 --------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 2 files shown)


yurishkuro:
I've noticed it long time ago, sorry thought I opened a ticket.

LGTM.

abhinav:
:+1:

breerly:
:shipit:

junchaowu:
Shall we add 'as' as well?

blampe:
Not as much value in that one afaik, it's basically always thrift.

PR #384, opened 2016-03-17
queue: Simpler thread-safe queue implementation

abhinav:
This implementation does not support `terminate()` or `put` backpressure, but it's a
much simpler implementation and easier to make thread-safe.

(Note: this PR is against the queue-terminate branch.)

### DIFF
--- a/tests/test_queue.py
+++ b/tests/test_queue.py
@@ -22,18 +22,13 @@ from __future__ import (
     absolute_import, unicode_literals, division, print_function
 )
 
-import sys
 import pytest
-from datetime import timedelta
-from six.moves import range
+import threading
+from collections import deque, defaultdict
 from tornado import gen
+from tornado.ioloop import IOLoop
 
-from tchannel._queue import Queue, QueueEmpty, TimeoutError
-
-
-def assert_invariant(q):
-    assert not (q._getters and (q._surplus or q._putters))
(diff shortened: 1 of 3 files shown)


abhinav:
CC @blampe @breerly @junchaowu @kriskowal

breerly:
Hard to complain - surprising how clean that impl is. :100:

tests/test_queue.py
```diff
@@ -42,241 +37,176 @@ def items():
...
-        assert_invariant(queue)
 
     got = yield got_futures
     assert got == items
 
 
 @pytest.mark.gen_test
-def test_unbounded_expired_gets_then_get_first():
-    queue = Queue()
+@pytest.mark.concurrency_test
```
blampe: (applause)

How much does it suck to run this locally? :)

blampe:
This is really impressive. If tests pass I trust it. I don't have any concerns except for the unboundedness, but we've currently got a bounded queue and we still might be seeing some memory leaks soooo this is still probably an improvement.

tests/test_queue.py
abhinav: Not a lot. It takes a couple seconds. I wanted to be able to skip that when I was verifying queue functionality.

kriskowal:
:ok:

PR #387, opened 2016-03-17
Log all response write errors except StreamClosedErorrs

abhinav:
This fixes yet another "Future exception was never retrieved" issue.


blampe:
:ship:

PR #388, opened 2016-03-18
Lower "cannot register on sync TChannel" to info.

abhinav:


junchaowu:
All right if we decide to lower this.

blampe:
ok

PR #390, opened 2016-03-23
Calls should reraise with the origianl traceback when retries run out

abhinav:
`raise error` eliminates the stack trace from the original failure.

This is being merged into the queue-terminate branch rather than master.

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -22,6 +22,7 @@ from __future__ import (
     absolute_import, division, print_function, unicode_literals
 )
 
+import sys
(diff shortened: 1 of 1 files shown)


abhinav:
@blampe @breerly @junchaowu

blampe:
> This is being merged into the queue-terminate branch rather than master.

Why not patch on master?

abhinav:
I'll cherry pick it on master but I want this change on the queue-terminate branch because we have some people test driving that branch right now.

blampe:
Gotcha sounds good! :ship:

PR #391, opened 2016-03-24
improve retry_limit docs

blampe:
@abhinav @breerly @junchaowu


blampe:
Landing this to address some questions that came up around this behavior, will follow-up with any necessary fixes.

abhinav:
:+1:

PR #393, opened 2016-03-28
Cherry pick heap-based peer selection into master

abhinav:
We want to release these changes without most of the breaking changes in the `mash` release branch. @junchaowu, can you take a close look at this since you're the primary author of this change? Especially, make sure I didn't miss anything in the cherry-pick.

CC @blampe @breerly

### DIFF
--- a/benchmarks/test_concurrent_requests.py
+++ b/benchmarks/test_concurrent_requests.py
@@ -1,0 +1,99 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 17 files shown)


tchannel/tornado/connection.py
```diff
@@ -122,7 +122,13 @@ def __init__(self, connection, tchannel=None, direction=None):
...
 
         # Map from message ID to futures for responses of outgoing calls.
-        self._outstanding = {}
+        self._outbound_pending_call = {}
+
+        # Map from message ID to outgoing request that is being sent.
+        self._outbound_pending_req = {}
+
+        # Map from message ID to outgoing response that is being sent.
+        self._outbound_pending_res = {}
```
abhinav: It appears that outbound pending req and res don't actually need references to the actual request and response objects. They can be sets of request IDs, or even just integers maintaining count on the number of ongoing requests and responses. @junchaowu can you confirm this?

junchaowu: I can say this is true in the future. But for now yes, it only needs the message ID to be stored as a tag.

benchmarks/test_concurrent_requests.py
```diff
@@ -0,0 +1,99 @@
...
+
+from tornado import gen, ioloop
+from tchannel import TChannel
+
+
+servers = None
+client = None
+
+
+def setUpServer(num):
```
blampe: Snake case :)

benchmarks/test_concurrent_requests.py
```diff
@@ -0,0 +1,99 @@
...
+        def hello(request):
+            return 'hello'
+        server.listen()
+        servers.append(server)
+     
+    return servers
+
+
+@gen.coroutine
+def setUpClient(servers):
```
blampe: Snake case

benchmarks/test_concurrent_requests.py
```diff
@@ -0,0 +1,99 @@
...
+            body='hi',
+            hostport=client.hostport,
+        )
+
+    raise gen.Return(client)
+
+
+@gen.coroutine
+def peer_test():
+    global servers, client
```
blampe: You don't use servers here. Also why the globals?

blampe:
Looks ok to me.

junchaowu:
it looks okay to me now.

PR #395, opened 2016-03-29
Reduce advertise interval frequency to once every 3 minutes

blampe:
@Raynos @abhinav

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Changes by Version
   TimeoutErrors.
 - Attempting to register endpoints against a synchronous TChannel now logs an
   INFO level message.
+- Reduced default advertisement interval to 3 minutes.
 
 
 0.21.10 (2016-03-17)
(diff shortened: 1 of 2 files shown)


abhinav:
More context? Why?

blampe:
This is something Jake suggested to take some load off of HB.

breerly:
Doesn't HB prune services in less than 3 mins? Or do they never actually fall off?

junchaowu:
3 minutes . that's a long time.

junchaowu:
Probably change the title from reduce to increase.

blampe:
@breerly @Raynos is going to coordinate a separate change for HB to go along with this.

Raynos:
@breerly the pruner runs every 5 mins; so thats a 1:5 ratio atm, ill change it to 9 minutes, which means we reduce to a 3:9 ratio.

abhinav:
:ok:

PR #396, opened 2016-03-30
PeerHeap: Make things cheaper

abhinav:
This adds minor improvements to the peer selection code that makes the change cheaper:

- Track number of pending outbound requests and responses as an integer rather than two dictionaries.
- Don't allocate new lists of connections when recalculating rank.

CC @junchaowu @blampe @breerly 

----

This is a PR against the peer-selection-cherrypick branch (#393).

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -68,8 +68,8 @@ class Peer(object):
         'order',
         'chosen_count',
         'on_conn_change',
+        'connections',
 
-        '_connections',
         '_connecting',
         '_on_conn_change_cb',
     )
@@ -88,7 +88,8 @@ class Peer(object):
             Host-port this Peer is for.
         :param rank:
             The rank of a peer will affect the chance that the peer gets
-            selected when the client sends outbound requests.
+            selected when the client sends outbound requests. Lower rank is
+            better.
         :param on_conn_change:
             A callback method takes Peer object as input and is called whenever
             there are connection changes in the peer.
@@ -102,7 +103,7 @@ class Peer(object):
         #: Collection of all connections for this Peer. Incoming connections
         #: are added to the left side of the deque and outgoing connections to
         #: the right side.
-        self._connections = deque()
+        self.connections = deque()
 
         # This contains a future to the TornadoConnection if we're already in
         # the process of making an outgoing connection to the peer. This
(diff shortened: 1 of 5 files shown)


junchaowu:
lgtm  this is cheaper.

blampe:
Nice! Ship if green.

breerly:
Can you post some numbers?

tchannel/tornado/peer.py
```diff
@@ -102,7 +103,7 @@ def __init__(self, tchannel, hostport, rank=None, on_conn_change=None):
         #: Collection of all connections for this Peer. Incoming connections
         #: are added to the left side of the deque and outgoing connections to
         #: the right side.
-        self._connections = deque()
+        self.connections = deque()
```
breerly: why make this public?

abhinav:
The performance improvement from this won't be that significant although I imagine memory usage will be better.

Before:

```
---------------------------------------------- benchmark: 1 tests ---------------------------------------------
Name (time in ms)          Min       Max      Mean   StdDev    Median      IQR  Outliers(*)  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------
test_peer_heap        182.1849  230.4020  206.2757  18.4516  218.5299  35.8012         11;0      25           1
---------------------------------------------------------------------------------------------------------------
```

After:

```
---------------------------------------------- benchmark: 1 tests ---------------------------------------------
Name (time in ms)          Min       Max      Mean   StdDev    Median      IQR  Outliers(*)  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------
test_peer_heap        178.7651  231.9150  201.8369  17.5618  211.8850  33.7656          9;0      25           1
---------------------------------------------------------------------------------------------------------------
```

tchannel/tornado/peer.py
abhinav: Because the `@property` version of this was `list(self._connections)` which can be a little expensive. The callers don't mutate or attempt to mutate this. In fact, the only place this is accessed externally is in PeerStrategy, which uses it to check if the peer even has any connections.

abhinav:
Updated

tchannel/tornado/peer.py
breerly: :+1:

breerly:
:shipit:

PR #400, opened 2016-04-06
make UnexpectedError a little more debuggable

blampe:
@abhinav

### DIFF
--- a/tchannel/tornado/dispatch.py
+++ b/tchannel/tornado/dispatch.py
@@ -21,6 +21,7 @@
 from __future__ import absolute_import
 
 import logging
+import sys
 from collections import namedtuple
 
 import tornado
@@ -233,17 +234,31 @@ class RequestDispatcher(object):
             e.id = request.id
             connection.send_error(e)
         except Exception as e:
+            exc_type, exc_obj, exc_tb = sys.exc_info()
+
+            # Walk to the TB to find our offending line.
+            while exc_tb.tb_next != None:
+                exc_tb = exc_tb.tb_next
+
+            description = "%r from %s in %s:%s" % (
+                e,
+                request.endpoint,
+                exc_tb.tb_frame.f_code.co_filename,
+                exc_tb.tb_lineno,
+            )
             error = UnexpectedError(
-                description="Unexpected Error: '%s'" % e,
+                description=description,
                 id=request.id,
                 tracing=request.tracing,
             )
-            response.set_exception(error)
+
+            # Grab exc_info() again because we stomped our tb above.
+            response.set_exception(error, exc_info=sys.exc_info())
             connection.request_message_factory.remove_buffer(response.id)
 
             connection.send_error(error)
             tchannel.event_emitter.fire(EventType.on_exception, request, error)
-            log.exception(error.description)
+            log.exception(e)
 
         raise gen.Return(response)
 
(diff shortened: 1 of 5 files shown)


tchannel/tornado/dispatch.py
```diff
@@ -233,17 +234,31 @@ def _on_post_response(future):
             e.id = request.id
             connection.send_error(e)
         except Exception as e:
+            exc_type, exc_obj, exc_tb = sys.exc_info()
```
abhinav: This will leave a reference cycle here. Add a try-finally block around the rest of the code and a 'del exc_tb' in the `finally` block.

Related warning in `sys.exc_info()` docs: https://docs.python.org/2/library/sys.html#sys.exc_info

tchannel/tornado/dispatch.py
```diff
@@ -233,17 +234,31 @@ def _on_post_response(future):
...
-            response.set_exception(error)
+
+            # Grab exc_info() again because we stomped our tb above.
+            response.set_exception(error, exc_info=sys.exc_info())
             connection.request_message_factory.remove_buffer(response.id)
 
             connection.send_error(error)
             tchannel.event_emitter.fire(EventType.on_exception, request, error)
-            log.exception(error.description)
+            log.exception(e)
```
abhinav: Make this a `log.error("some message", exc_info=exc_info)`?

tchannel/tornado/dispatch.py
blampe: Great catch. Is it a problem that I re-assign `exc_tb`?

tchannel/tornado/dispatch.py
```diff
@@ -233,17 +234,36 @@ def _on_post_response(future):
...
         except Exception as e:
-            error = UnexpectedError(
-                description="Unexpected Error: '%s'" % e,
-                id=request.id,
-                tracing=request.tracing,
-            )
-            response.set_exception(error)
-            connection.request_message_factory.remove_buffer(response.id)
+            try:
+                exc_type, exc_obj, exc_tb = sys.exc_info()
```
abhinav: Nit: Generally, the assignment should take place outside the try block if you're using the value in the finally block.

    x = foo
    try:
        # ...
    finally:
        do_stuff_with(x)

That's because if it's inside the try block, and the block fails before assigning, the variable will be undefined inside the finally block. This doesn't apply as strongly here because `sys.exc_info()` will never fail, but it still makes sense to assign outside the block.

tchannel/tornado/dispatch.py
```diff
@@ -233,17 +234,36 @@ def _on_post_response(future):
...
-            connection.send_error(error)
-            tchannel.event_emitter.fire(EventType.on_exception, request, error)
-            log.exception(error.description)
+                # Grab exc_info() again because we stomped our tb above.
+                response.set_exception(error, exc_info=sys.exc_info())
+                connection.request_message_factory.remove_buffer(response.id)
+
+                connection.send_error(error)
+                tchannel.event_emitter.fire(EventType.on_exception, request, error)
+                log.error("Unexpected error", exc_info=sys.exc_info())
```
abhinav: You want to use the original output of `sys.exc_info()` here. If any of the functions called above throw or catch exceptions (even as `@coroutine`s), `sys.exc_info()` will be overwritten.

tchannel/tornado/dispatch.py
abhinav: Re-assigning shouldn't be a problem because then you're no longer holding onto the reference, but since you want to log the original *full* stack trace in `log.error`, you may need to hold onto the original reference walk the traceback in a separate variable.

abhinav:
Rebased on master and fixed typos in comments.

LGTM

PR #401, opened 2016-04-06
Add a timeout to the VCR proxy call

blampe:
@abhinav (shrug)

### DIFF
--- a/tchannel/testing/vcr/patch.py
+++ b/tchannel/testing/vcr/patch.py
@@ -103,6 +103,7 @@ class PatchedClientOperation(object):
             vcr_response_future = tchannel.thrift(
                 proxy.VCRProxy.send(vcr_request),
                 hostport=self.vcr_hostport,
+                timeout=1.0,
             )
 
         try:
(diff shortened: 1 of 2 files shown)


abhinav:
:+1:

tchannel/testing/vcr/patch.py
```diff
@@ -103,6 +103,7 @@ def send(self, arg1, arg2, arg3,
             vcr_response_future = tchannel.thrift(
                 proxy.VCRProxy.send(vcr_request),
                 hostport=self.vcr_hostport,
+                timeout=1.0,
```
breerly: - you'll want this to be configurable by those using vcr
- seems like a low default

abhinav: Oh, you're right. If a recording does not exist, a real request will be made so we should use the timeout from the original request + some delay for the proxy request.

abhinav: Created https://github.com/uber/tchannel-python/pull/403

PR #403, opened 2016-04-06
vcr: Use ttl of the original request for VCR proxy requests

abhinav:
Rather than using a hard-coded one second timeout, we should use whatever
the user specified for that request.

(I also made the error message for NoPeersAvailableError more informative.)


abhinav:
CC @breerly @blampe @junchaowu

blampe:
:sheep:

breerly:
oh, cool :ship:

PR #404, opened 2016-04-06
Remove a warning for unconsumed errors

blampe:
@abhinav

### DIFF
--- a/tchannel/tornado/message_factory.py
+++ b/tchannel/tornado/message_factory.py
@@ -317,7 +317,7 @@ class MessageFactory(object):
         elif message.message_type == Types.ERROR:
             context = self.message_buffer.pop(message.id, None)
             if context is None:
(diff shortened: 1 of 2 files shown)


tchannel/tornado/message_factory.py
```diff
@@ -317,7 +317,7 @@ def build(self, message):
         elif message.message_type == Types.ERROR:
             context = self.message_buffer.pop(message.id, None)
             if context is None:
-                log.warn('Unconsumed error %s', context)
```
abhinav: Actually, this means that we don't have a request to send this error to. We should log the error regardless.  Maybe change this to `log.info(...)` and log the `message` rather than the `context`?

blampe: sure

abhinav:
:+1:

PR #409, opened 2016-04-14
constraint to tornado 4.3

breerly:
@abhinav @blampe @junchaowu


abhinav:
ok

thecadams:
thanks so much for the quick response here.

breerly:
@thecadams no problem - it's a bummer that pip doesn't respect version library-declared version constraints. I think we'll probably end up in a place where we do a hard fail.

abhinav:
LGTM

PR #411, opened 2016-04-19
Expose host, port, close(), and is_closed()

breerly:
This catches up the public `TChannel` class w `tchannel.tornado.tchannel.TChannel`, which should stop us from reaching into `tchannel._dep_tchannel` from the outside.


abhinav:
👍

junchaowu:
👍

PR #412, opened 2016-06-07
Support OpenTracing

yurishkuro:
* Add OpenTracing hook
* Introduce RequestContextProvider to allow pluggable StackContext propagation mechanism
* Apply a patch to tornado.testing bug when 'localhost' resolves to >1 IP
* Fix 'cannot import name is_list_like' by bounding pygal

@abhinav @blampe @breerly

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,9 @@ setup(
         # tchannel deps
         'thriftrw>=0.4,<2',
         'threadloop>=1,<2',
+
+        # tracing deps
+        'opentracing>=1.0rc4,<2'
     ],
     extras_require={
         'vcr': ['PyYAML', 'mock', 'wrapt'],
--- a/tchannel/context.py
+++ b/tchannel/context.py
@@ -73,12 +73,25 @@ class RequestContext(object):
         _LOCAL.context = self._old_context
 
 
-def get_current_context():
-    """
-    :return: The current :py:class:`RequestContext` for this thread.
-    """
-    return _LOCAL.context
+# noinspection PyMethodMayBeStatic
+class RequestContextProvider(object):
+    def get_current_context(self):
+        """
+        :return: The current :py:class:`RequestContext` for this thread.
+        """
+        return _LOCAL.context
 
+    def request_context(self, parent_tracing):
+        """
+        Factory method meant to be used as:
 
-def request_context(parent_tracing):
-    return StackContext(lambda: RequestContext(parent_tracing))
+        .. code-block:: python
+
+            with tchannel.context_provider.request_context(parent_tracing):
+                handler_fn()
+
+        :param parent_tracing:
+        :return:
+        """
+        # TODO should this be using a thread-safe version of StackContext?
+        return StackContext(lambda: RequestContext(parent_tracing))
--- a/tchannel/tornado/tchannel.py
+++ b/tchannel/tornado/tchannel.py
@@ -34,6 +34,7 @@ import tornado.tcpserver
 from tornado.netutil import bind_sockets
 
 from . import hyperbahn
+from ..context import RequestContextProvider
 from ..deprecate import deprecate
 from ..enum import enum
 from ..errors import AlreadyListeningError
@@ -74,7 +75,8 @@ class TChannel(object):
 
     def __init__(self, name, hostport=None, process_name=None,
                  known_peers=None, trace=False, dispatcher=None,
-                 reuse_port=False, _from_new_api=False):
+                 reuse_port=False, context_provider_fn=None,
+                 _from_new_api=False):
         """Build or re-use a TChannel.
 
         :param name:
@@ -98,12 +100,23 @@ class TChannel(object):
         :param trace:
             Flag to turn on/off zipkin trace. It can be a bool variable or
             a function that return true or false.
+
+        :param context_provider_fn:
+            A getter function to retrieve an instance of
+            context.RequestContextProvider used to manage
+            thread-local request context.
         """
 
         self._state = State.ready
+        if context_provider_fn:
+            self.context_provider_fn = context_provider_fn
+        else:
+            context_provider = RequestContextProvider()
+            self.context_provider_fn = lambda: context_provider
 
         if not dispatcher:
-            self._handler = RequestDispatcher()
+            self._handler = RequestDispatcher(
+                context_provider_fn=lambda: self.context_provider_fn())
         else:
             self._handler = dispatcher
 
--- a/tchannel/tracing/opentracing.py
+++ b/tchannel/tracing/opentracing.py
@@ -38,7 +38,7 @@ class SpanWrapper(tchannel.zipkin.trace.Trace):
     integrate with non-TChannel RPC calls.
     """
     def __init__(self, span, *args, **kwargs):
-        super(SpanWrapper, self).__init__(args, *kwargs)
+        super(SpanWrapper, self).__init__(*args, **kwargs)
         self.span = span
 
     @staticmethod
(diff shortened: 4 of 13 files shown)


blampe:
Hm Travis didn't run, I think because it was disabled for PRs. I've re-enabled it so the next time you push to your branch it should run.

tests/test_opentracing_hook.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def get_current_context(self):
+        """Returns a Context that has parent_tracing attribute"""
+        context = RequestContextManager.current_context()
+        if not context:
+            return None
+        if not hasattr(context, 'parent_tracing'):
+            context = MockRequestContext(
+                span=context.span,
+                parent_tracing=SpanWrapper(span=context.span)
```
abhinav: Is this supposed to be `context.parent_tracing`?

abhinav:
The overall approach looks good to me.

tests/test_opentracing_hook.py
yurishkuro: @abhinav not sure what you mean. The primary (opentracing) context does not have a `parent_tracing` attribute, but tchannel expects it from `get_current_context` call, so I replace the opentracing context with a tchannel-compatible extension.

tchannel/context.py
```diff
@@ -73,12 +73,25 @@ def __exit__(self, type, value, traceback):
...
-def get_current_context():
-    """
-    :return: The current :py:class:`RequestContext` for this thread.
-    """
-    return _LOCAL.context
-
-
-def request_context(parent_tracing):
-    return StackContext(lambda: RequestContext(parent_tracing))
+# noinspection PyMethodMayBeStatic
```
yurishkuro: @abhinav @blampe any concerns that I am removing these static methods?  I made sure tchannel code does not call them anymore, it always uses context provider instance.

tests/test_opentracing_hook.py
abhinav: 👍

tchannel/context.py
abhinav: No, this is fine. Converting the top-level functions into an object is a clean enough way of making this pluggable.

setup.py
```diff
@@ -55,6 +55,9 @@
         # tchannel deps
         'thriftrw>=0.4,<2',
         'threadloop>=1,<2',
+
+        # tracing deps
+        'opentracing>=1.0rc4'
```
breerly: Please put an upper bound. Suggest:

```
opentracing>=1,<2'
```

tests/test_opentracing_hook.py
```diff
@@ -0,0 +1,467 @@
...
+                span.finish()
+
+
+@pytest.mark.parametrize('endpoint,transport,expect_spans', [
+    ('endpoint1', 'tchannel', 5),  # tchannel->tchannel
+    ('endpoint3', 'tchannel', 5),  # tchannel->http
+    ('/', 'http', 5),  # http->tchannel
+])
+@pytest.mark.gen_test
+def test_trace_mixed(endpoint, transport, expect_spans,
```
breerly: nice test!

tchannel/tornado/tchannel.py
```diff
@@ -74,7 +75,8 @@ class TChannel(object):
 
     def __init__(self, name, hostport=None, process_name=None,
                  known_peers=None, trace=False, dispatcher=None,
-                 reuse_port=False, _from_new_api=False):
+                 reuse_port=False, _from_new_api=False,
+                 context_provider_fn=None):
```
breerly: put this arg ahead of the private `_from_new_api` flag plz

tchannel/tracing/opentracing_hook.py
```diff
@@ -0,0 +1,187 @@
...
+
+import logging
+
+import opentracing
+import tchannel.event
+import tchannel.zipkin.trace
+from opentracing.ext import tags
+from opentracing.propagation import Format
+
+
```
breerly: suggest calling this file just `opentracing.py`

setup.py
yurishkuro: I made it `opentracing>=1.0rc4,<2`, as pip doesn't like `>=1' when there hasn't been an official OT release, only RCs.

blampe:
Have you tried testing this out on something like Larry? It looks OK to me but this code path has had a few problems with memory leaks that I want to make sure we catch up-front before releasing (this would be a minor bump, regardless).

yurishkuro:
I can integrate it into Larry, but we'd first need to land this. Adding an optional hook doesn't introduce any memory leaks by itself.

blampe:
We can actually install the branch from GH (before landing) by changing the
requirement from "tchannel" to "git+git://
github.com/yurishkuro/tchannel-python.git@opentracing-integration. That way
if you need to do any tweaks here you can test them immediately without
needing another PR/land cycle.

yurishkuro:
@blampe @breerly tested with "larry", should be ready to land if no more comments

breerly:
lgtm!

tchannel/tracing/opentracing.py
```diff
@@ -0,0 +1,217 @@
...
+            span_id, and parent_span_id, where values are all i64.
+        :param log_exception_fn:
+            Zero-arg function used to log exceptions. Defaults to
+            logging.exception('OpenTracing exception').
+        """
+        self.context_provider = context_provider
+        self._tracer = tracer
+        self.span_to_trace_fn = span_to_trace_fn
+        self.log_exception_fn = log_exception_fn if log_exception_fn \
+            else lambda: logging.exception('OpenTracing exception')
```
abhinav: Nit: Use full if-else blocks here.

```python
if not log_exception_fn:
  log_exception_fn = lambda: logging.exception(..)
self.log_exception_fn = log_exception_fn
```

tchannel/tracing/opentracing.py
```diff
@@ -0,0 +1,217 @@
...
+        if span:
+            return span
+        parent_tracing = getattr(context, 'parent_tracing', None)
+        if not parent_tracing:
+            return None
+        return getattr(parent_tracing, 'span', None)
+
+    def _span_to_trace(self, span):
+        try:
+            ids = self.span_to_trace_fn(span) if self.span_to_trace_fn else {}
```
abhinav: Nit: Prefer if blocks

```python
ids = {}
if self.span_to_trace_fn:
  ids = self.span_to_trace_fn(span)
```

tests/test_opentracing.py
```diff
@@ -0,0 +1,517 @@
...
+import tornado
+import tornado.gen
+import tornado.testing
+import tornado.web
+from basictracer import BasicTracer, SpanRecorder
+from opentracing import Format
+from opentracing_instrumentation.client_hooks.tornado_http import \
+    install_patches, reset_patchers
+from opentracing_instrumentation.request_context import \
+    RequestContext, RequestContextManager, \
```
abhinav: Nit: Prefer parens over `\`

```python
from opentracing_instrumentation.request_context import (
  RequestContext,
  RequestContextManager,
  span_in_stack_context,
  get_current_span,
)
```

tests/test_opentracing.py
```diff
@@ -0,0 +1,517 @@
...
+from tornado.stack_context import StackContext
+
+BAGGAGE_KEY = b'baggage'
+
+
+# It seems under Travis the 'localhost' is bound to more then one IP address,
+# which triggers a bug: https://github.com/tornadoweb/tornado/issues/1573
+# Here we monkey-patch tornado/testing.py with the fix from
+#   https://github.com/tornadoweb/tornado/pull/1574
+def patched_bind_unused_port(reuse_port=False):
```
abhinav: lol

tests/test_opentracing.py
```diff
@@ -0,0 +1,517 @@
...
+        context = MockRequestContext(span=span, parent_tracing=parent_tracing)
+        # TODO should this be using a thread-safe version of StackContext?
+        return StackContext(lambda: RequestContextManager(context=context))
+
+
+@pytest.yield_fixture
+def http_patchers():
+    install_patches.__original_func()  # install_patches func is a @singleton
+    try:
+        yield 'http_patchers'
```
abhinav: You don't have to yield a fake value if you don't have a value to provide. Just a `yield` statement is enough.

tchannel/tracing/opentracing.py
yurishkuro: if I do, I get "PEP 8 do not assign lambda, use def"

tchannel/tracing/opentracing.py
yurishkuro: isn't it less efficient? you need to allocate an empty dictionary.

tests/test_opentracing.py
```diff
@@ -0,0 +1,517 @@
...
+    finally:
+        reset_patchers()
+
+
+def register(tchannel, http_client=None, base_url=None):
+    @tchannel.raw.register('endpoint1')
+    @tornado.gen.coroutine
+    def handler1(request):
+        print 'handler1 hit'
```
abhinav: Did you mean to delete these print statementsr?

yurishkuro: they are quite helpful if the test is mis-behaving. When running `make test` these do not show up as long as the tests are successful, so is there any harm in keeping them?

abhinav: Mostly just the noise. We try not to have random print statements in tests or code.

(Also, they will show up if you run tests with `-s`, which you need if you're using `pdb.set_trace()`.)

tchannel/tracing/opentracing.py
abhinav: I don't think this will have any measurable impact on performance.

abhinav:
LGTM

PR #414, opened 2016-07-04
Fix opentracing dependency to 1.0rc4 in anticipation of API changes

yurishkuro:
There is a [large API change](https://github.com/opentracing/opentracing.github.io/pull/100) coming soon to OpenTracing, so it's best to pin the API version to RC4 to avoid breaking people.


yurishkuro:
The next OT version will be 2.x, so this is not needed

PR #415, opened 2016-07-05
Integrate with OpenTracing 1.1 API

yurishkuro:
Ready for review. Main changes:
 * native tchannel tracing removed completely
 * `tracing` field is still the same, and is initialized from the OpenTracing Span if the tracer supports Zipkin format, otherwise it's initialized randomly
 * the actual tracing is done via OpenTracing API
 * Baggage propagation only works with JSON/Thrift encoding as it needs app headers to propagate. Plain tracing works with Raw encoding if tracer is Zipkin compatible
 * `RequestContext` is removed and replaced by pluggable `TracingContextProvider`, which defaults to `opentracing_instrumentation` StackContext+thread-local based implementation, for interoperability with http patching
 * crossdock-based tests in PR #423


tchannel/tchannel.py
```diff
@@ -25,20 +25,21 @@
 import json
 import logging
 
+from opentracing.ext import tags as ext_tags
```
yurishkuro: TODO: move to ServerTracer

tchannel/tchannel.py
```diff
@@ -181,15 +186,20 @@ def call(
             retry_limit = retry.DEFAULT_RETRY_LIMIT
 
         # TODO - allow filters/steps for serialization, tracing, etc...
-        context = self.context_provider.get_current_context()
+
+        if trace is None:
+            trace = self._dep_tchannel.trace
+        trace = trace() if callable(trace) else trace
+        if not trace and tracing_span:
+            tracing_span.set_tag(ext_tags.SAMPLING_PRIORITY, 0)
```
yurishkuro: use apply_trace_flag()

blampe:
Can we leave some wrappers to preserve `RequestContext` and anything else getting removed here? I'd prefer to have this be a backwards-compatible change, otherwise it'll be tricky to roll out.

yurishkuro:
> Can we leave some wrappers to preserve RequestContext and anything else getting removed here? I'd prefer to have this be a backwards-compatible change, otherwise it'll be tricky to roll out.

There were two issues with the original RequestContext:
 # in order to provide interoperability between say tchannel server endpoint and http downstream calls, there must be a single "request context" propagated through the app. I have abstracted it out in tchannel via `TracingRequestContext`, with default implementation falling back onto the context provided by `openrtacing_instrumentation`, which is used by monkey-patched http methods.
 # the original request context in tchannel was used _only_ to carry `parent_tracing` field, with no possible extensions, so it seemed reasonable to replace it with `TracingRequestContext`

@blampe what kind of difficulties do you foresee with rolling this out? Do you know of some other uses of the original request context?

blampe:
> what kind of difficulties do you foresee with rolling this out? Do you know of some other uses of the original request context?

I just want to make sure we're not breaking imports or interfaces (we've kinda committed to no breaking changes). If you're confident that no one else is using this except for the Jaeger client (and that Jaeger is compatible with this and the old tracing implementation) then we can go ahead without a compatibility shim. 

Also we should include a changelog entry under a new minor version.

yurishkuro:
@blampe I think you're right, I can add `context.py` and `ZipkinHook` back as noop implementations just for compatibility reasons. I would not expect anyone to use `context.py` because there's nothing interesting there to use, but better be safe. And ZipkinHook is definitely being used in our `clay-hyperbahn` lib, so I would make it a no-op.

tchannel/messages/common.py
```diff
@@ -66,6 +67,22 @@
...
 
 
+def random_tracing():
+    """
+    Create new Tracing() tuple with random IDs.
+    """
+
+    def _uniq_id():  # TODO: cython
+        """Create a random 64-bit unsigned long."""
+        return random.getrandbits(64)
```
abhinav: Why inline this instead of making it top-level?

yurishkuro: merely from encapsulation pov, as nobody else needs it. But if it's detrimental to performance, I can move it.

abhinav: This probably doesn't affect performance significantly, but it's not a common pattern in this codebase. The leading `_` in the name should be enough in discouraging its use from elsewhere. Please move it outside.

tchannel/thrift/client.py
```diff
@@ -30,6 +30,7 @@
 from tchannel.errors import OneWayNotSupportedError
 
 from ..serializer.thrift import ThriftSerializer
+from ..import tracing
```
abhinav: nit: space between `..` and `import`

tchannel/tracing.py
```diff
@@ -0,0 +1,206 @@
...
+
+log = logging.getLogger('tchannel')
+
+
+# noinspection PyMethodMayBeStatic
+class TracingContextProvider(object):
+    """Tracks the OpenTracing Span currently in effect.
+
+    Distributed tracing requires context propagation inside the application,
+    do that outbound (downstream) calls can be part of the same trace as the
```
abhinav: `s/do/so/`

tchannel/tracing.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        pass  # TODO(ys) implement start_basic_span
+
+    def start_span(self, request, headers, peer_host, peer_port):
+        """
+        Starts a new server-side span. If the span has already been started
+        by `start_basic_span`, this method only adds baggage from the headers.
+        :param request:
+        :param headers:
+        :return:
```
abhinav: Please document the args and the return type. This looks like `request` must always be `tchannel.tornado.Request` since `request.headers` is assumed to be the transport headers.

tests/test_context.py
```diff
@@ -18,55 +18,86 @@
...
-        arg2='req headers',
-        arg3='req body',
-        hostport=server.hostport,
-    )
-
-    assert context[0].parent_tracing.name == 'endpoint1'
-    assert context[1].parent_tracing.name == 'endpoint2'
+# from __future__ import absolute_import
+#
+# import threading
```
abhinav: What's up with this file?

tests/test_tracing.py
```diff
@@ -0,0 +1,539 @@
...
+from tornado import netutil
+from tornado.httpclient import HTTPRequest
+
+BAGGAGE_KEY = b'baggage'
+
+
+# It seems under Travis the 'localhost' is bound to more then one IP address,
+# which triggers a bug: https://github.com/tornadoweb/tornado/issues/1573
+# Here we monkey-patch tornado/testing.py with the fix from
+#   https://github.com/tornadoweb/tornado/pull/1574
```
abhinav: lulz

abhinav:
Core logic looks right.

tests/test_context.py
yurishkuro: I think it needs to be deleted. The new `context.py` no longer provides the same propagation features as before, its methods are only for API backwards compatibility.

tests/test_tracing.py
```diff
@@ -0,0 +1,539 @@
...
+    assert 1 == len(set([s.context.trace_id for s in spans])), \
+        'all spans must have the same trace_id'
+
+
+# @pytest.mark.parametrize('ex_cls,start_span', [
+#     (opentracing.UnsupportedFormatException, 0),
+#     (opentracing.InvalidCarrierException, 0),
+#     (opentracing.TraceCorruptedException, 1),
+# ])
+# def test_before_receive_request(ex_cls, start_span):
```
blampe: Were these meant to be enabled?

blampe:
Looks good from my end, thanks for taking care of the backwards compatibility. I trust that you've tested this and are comfortable rolling it out. 

Let's call this 0.27.0 in the changelog.

setup.py
```diff
@@ -56,8 +56,9 @@
         'thriftrw>=0.4,<2',
         'threadloop>=1,<2',
 
-        # tracing deps
-        'opentracing>=1.0rc4,<2'
+        # tracing deps - pinned due to upcoming API changes
+        'opentracing>=1.0rc4,<1.1',
+        'opentracing_instrumentation>=1.1,<1.2',
```
breerly: I'm still not sure why you're pinning to the minor here - are you guys planning breaking changes within a `1.x`?

blampe:
@yurishkuro let us know if this is ready for review or if you're still working on it.

yurishkuro:
@blampe there was one test failing, it's ready for review now.

tchannel/context.py
```diff
@@ -36,6 +36,11 @@ def __init__(self):
 class RequestContext(object):
     """Tracks the :py:class:`Request` currently being handled.
 
+    NOTE: this class is deprecated and not used by TChannel itself.
+    It is only kept for backwards compatibility.
+
+    See tracing.TracingContextProvider for replacement.
```
abhinav: Consider replacing with Sphinx's deprecated option:

```
.. deprecated:: 0.27.0
   Use tracing.TracingContextProvider
```

tchannel/context.py
```diff
@@ -75,6 +80,12 @@ def __exit__(self, type, value, traceback):
 
 # noinspection PyMethodMayBeStatic
 class RequestContextProvider(object):
+    """
+    NOTE: this class is deprecated and not used by TChannel itself.
+    It is only kept for backwards compatibility.
+
+    See tracing.TracingContextProvider for replacement.
```
abhinav: Same as above

tests/test_tracing.py
```diff
@@ -0,0 +1,523 @@
...
+        assert 1 == len(set([s.trace_id for s in spans])), \
+            'all spans must have the same trace_id'
+
+
+@pytest.mark.parametrize('encoding,operation', [
+    ('json', 'foo'),
+    ('thrift', 'X::thrift2'),
+])
+@pytest.mark.gen_test
+def __test_span_tags(encoding, operation, tracer, thrift_service):
```
abhinav: Why `__`?

abhinav:
What's up with the `__test_*` tests? Are those being skipped because they're expected to fail? You should probably use pytest's xfail or skip functionality for that.

    @pytest.mark.skip(reason='...')
    def test_foo():
        # ...

    @pytest.mark.xfail
    def test_bar():
        # ...

LGTM otherwise

tests/test_tracing.py
yurishkuro: oh, good catch, I did this to disable them temporarily when troubleshooting the failure in the other one.

Issue #416, opened 2016-07-07
VCR tests are unstable - socket is None no tornado.iostream


yurishkuro:
A number of different tests related to VCR are randomly breaking CI, for example

https://travis-ci.org/uber/tchannel-python/jobs/143115312

```
_____________ ERROR at teardown of test_unexpected_error[hostPort] _____________
cassette = <InstanceDouble of <class 'tchannel.testing.vcr.cassette.Cassette'> object at 0x2b3e61913290>
unpatch = <function reset at 0x2b3e61987d70>
    @pytest.yield_fixture
    def vcr_service(cassette, unpatch):
        with VCRProxyService(cassette, unpatch) as vcr_service:
>           yield vcr_service

tests/testing/vcr/test_server.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tchannel/testing/vcr/server.py:219: in __exit__
    self.stop()
tchannel/testing/vcr/server.py:205: in stop
    self.tchannel._dep_tchannel.close()
tchannel/tornado/tchannel.py:236: in close
    self.peers.clear()
tchannel/tornado/peer.py:581: in clear
    peer.close()
tchannel/tornado/peer.py:246: in close
    connection.close()
tchannel/tornado/connection.py:313: in close
    self.connection.close()
.tox/py27-tornado43/lib/python2.7/site-packages/tornado/iostream.py:436: in close
    self.close_fd()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <tornado.iostream.IOStream object at 0x2b3e619112d0>
    def close_fd(self):
>       self.socket.close()
E       AttributeError: 'NoneType' object has no attribute 'close'
```

breerly:
@yurishkuro did #417 solve this? I can close if so.

yurishkuro:
#417 masked the problem so that the tests don't fail spontaneously. Considering that the issue kept occurring in vcr tests, which according to @abhinav start a new thread, there is probably a race condition somewhere there, rather than a fundamental problem in Tornado.

PR #417, opened 2016-07-07
Protect against "socket is None" in iostream.close() with try/catch

yurishkuro:
Temporary fix for #416


blampe:
+@abhinav

abhinav:
:shipit:

abhinav:
Ignore the coveralls failure. We need to adjust the threshold.

breerly:
ty sir

Issue #419, opened 2016-07-08
Calling vcr.use_cassette reads/parses fixture file each time it's called


shubik22:
I work on a project that uses tchannel-python and decorates a large number of tests with vcr.use_cassette to stub out tchannel calls.  It uses the same yaml file as a fixture everywhere it calls vcr.use_cassette.  It's a somewhat large yaml file; on my MacBook Pro, it takes about half a second to load/parse the file.

The issue is that each time vcr.use_cassette is called, the yaml file is loaded and parsed anew, adding at least 0.5 seconds to each test in the case above.  See [here](https://github.com/uber/tchannel-python/blob/master/tchannel/testing/vcr/cassette.py#L170):

```
    def _load(self):
        try:
            with open(self.path, 'rb') as f:
                data = f.read()
                self.existed = True
        except IOError:
            return  # nothing to read

        data = yaml.load(data)
        if not (data and 'interactions' in data):
            return  # file was probably empty

        if int(data['version']) != VERSION:
            raise UnsupportedVersionError(
                'Cassette at "%s" is an unsupported version of the '
                'format: version %s' % (self.path, str(data['version']))
            )

        self._available = [
            Interaction.to_native(i) for i in data['interactions']
        ]
```

I'd propose adding an optional cache for interactions loaded from a yaml file, so that users who have a read-only file used throughout their test suite can read/parse it once (rather than every time vcr.use_cassette is called).

I don't have a lot of experience with python, but the code would probably look something like this.  There'd be a cache class:

```
class _InteractionsCache(object):

    _cache = {}

    @classmethod
    def save_interactions(cls, path, interactions):
        cls._cache[path] = interactions

    @classmethod
    def load_interactions(cls, path):
        try:
            return deepcopy(cls._cache[path])
        except KeyError:
            return None

```

(Do I need to work about any sort of thread-safety issues with this kind of cache in Python?)

Then I'd modify Cassette._load (additions are the blocks starting with `if self.use_cache` at beginning and end of the method):

```
    def _load(self):
        if self.use_cache:
            interactions = _InteractionsCache.load_interactions(self.path)
            if interactions:
                self._available = interactions
                return

        try:
            with open(self.path, 'rb') as f:
                data = f.read()
                self.existed = True
        except IOError:
            return  # nothing to read

        data = yaml.load(data)
        if not (data and 'interactions' in data):
            return  # file was probably empty

        if int(data['version']) != VERSION:
            raise UnsupportedVersionError(
[...]

PR #420, opened 2016-07-08
Add cache for cassette yaml files

shubik22:
See https://github.com/uber/tchannel-python/issues/419

### DIFF
--- a/tchannel/testing/vcr/cassette.py
+++ b/tchannel/testing/vcr/cassette.py
@@ -102,6 +102,9 @@ class Cassette(object):
             If specified, this is a collection of matcher names. These
             specify which attributes on two requests should match for them to
             be considered equal.
+        :param serializer:
+            An object with a ``dump(obj)`` and a ``load(str)`` method used to
+            serialize and deserialize the cassette.
         """
         # TODO move documentation around
         record_mode = record_mode or RecordMode.ONCE
@@ -220,7 +223,7 @@ class Cassette(object):
             interactions.extend(self._available)
         interactions.extend(self._recorded)
 
-        data = yaml.dump(
+        data = self.serializer.dump(
             {
                 'interactions': [i.to_primitive() for i in interactions],
                 'version': VERSION,
--- a/tchannel/testing/vcr/cassette.py
+++ b/tchannel/testing/vcr/cassette.py
@@ -25,6 +25,8 @@ import os.path
 from itertools import chain
 from collections import deque
 from collections import namedtuple
+from copy import deepcopy
+from hashlib import sha256
 
 from . import yaml
 from .exceptions import RequestNotFoundError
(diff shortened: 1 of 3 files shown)


tchannel/testing/vcr/cassette.py
```diff
@@ -110,6 +114,10 @@ def __init__(self, path, record_mode=None, matchers=None):
         # this was a new cassette and the YAML file did not exist.
         self.existed = False
 
+        if serializer is None:
+            serializer = yaml
+        self.serializer = serializer
```
abhinav: Can you add this to the docstring above?

```
:param serializer:
  An object with a ``dump(obj)`` and a ``load(str)`` method used to
  serialize and deserialize the cassette.
```

tchannel/testing/vcr/cassette.py
```diff
@@ -168,14 +177,19 @@ def _find(self, request, haystack=None):
...
                 self.existed = True
+                file_hash = sha256(data).hexdigest()
+                if (self.path, file_hash) in self._cache:
+                    self._available = deepcopy(self._cache[(self.path, file_hash)])
+                    return
         except IOError:
             return  # nothing to read
 
-        data = yaml.load(data)
+        data = self.serializer.load(data)
```
abhinav: You need to update all calls to `yaml.dump` as well.

tchannel/testing/vcr/cassette.py
```diff
@@ -168,14 +177,19 @@ def _find(self, request, haystack=None):
...
 
     def _load(self):
+        file_hash = None
         try:
             with open(self.path, 'rb') as f:
                 data = f.read()
                 self.existed = True
+                file_hash = sha256(data).hexdigest()
+                if (self.path, file_hash) in self._cache:
+                    self._available = deepcopy(self._cache[(self.path, file_hash)])
```
abhinav: Nit: This can be simplified to,

```python
cached = self._cache.get((self.path, file_hash))
if cached is not None:
  self._available = deepcopy(cached)
  return
```

shubik22: Sweet, fixed.

tchannel/testing/vcr/cassette.py
shubik22: Done.

tchannel/testing/vcr/cassette.py
shubik22: Good catch, added.

tchannel/testing/vcr/cassette.py
```diff
@@ -168,14 +180,20 @@ def _find(self, request, haystack=None):
...
     def _load(self):
+        file_hash = None
         try:
             with open(self.path, 'rb') as f:
                 data = f.read()
                 self.existed = True
+                file_hash = sha256(data).hexdigest()
+                if (self.path, file_hash) in self._cache:
+                    self._available = \
+                        deepcopy(self._cache[(self.path, file_hash)])
```
abhinav: Nit: Avoid using `\`. Parens are cleaner.

```python
self._available = deepcopy(
    self._cache[(self.path, file_hash)]
)
```

(Although this shouldn't be a problem if you're using the `self._cache.get(...)` approach instead.)

abhinav:
LGTM with a minor nit. CC @breerly and @blampe if they want to have a look.

We can merge and release this on Monday.

blampe:
Nice, thanks for the PR! 🚢

PR #428, opened 2016-08-10
Set Trace.parent_id to 0 if it's None

### LINKED ISSUE #427: People reporting timeout issues with v.0.27 when advertising to Hyperbahn
I was able to reproduce it on the test server and tracked it down to Jaeger tracer returning `Trace.parent_id == None` instead of 0. So the fix is obvious and I confirmed it solves the issue, but it's sad that none of the existing unit tests have caught it.

yurishkuro:
Fixes #427

@abhinav @blampe let's release it asap as it's breaking people who upgrade to 0.27, and then think how to add tests catching this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.27.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Bug fix: set Trace.parent_id to 0 if it's None
 
 
 0.27.0 (2016-08-08)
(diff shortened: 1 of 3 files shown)


blampe:
LGTM, @abhinav will handle the release this morning.

blampe:
This might be a good crossdock test case in case we're not already covering it -- issue an un-instrumented request to every language and assert some traces come out.

yurishkuro:
@blampe I think that's better as a unit test. Crossdock is best at testing interop, but this scenario is a lack of interop, which is just as easy to simulate in a unit test, and actually easier to verify that a trace is started, since unit tests have access to the trace reporter in memory.

Note that the actual issue fixed here is different. From what I gathered the ad request was never even leaving the server. Perhaps parent_id = None was causing some serialization issue, which wasn't surfacing via any exception, and later the Future was timing out.

abhinav:
Add an entry to the changelog and :shipit:

PR #429, opened 2016-08-11
Use latest Go crossdock build

yurishkuro:
Have to use 'dev' instead of 'latest' since python-go does not update master branch often


abhinav:
👍

Issue #431, opened 2016-08-16
TChannel client increases ad frequency when it fails


kriskowal:
Instead of backing off, TChannel node’s Hyperbahn ad loop increases ad frequency linearly when it fails.

This is caused by scheduling a retry before waiting for a success or failure.

breerly:
From the description, I get the impression that this issue should have been created on github.com/uber/tchannel-node... no?

blampe:
I think @kriskowal was describing the expected behavior, which Python doesn't do :(

PR #432, opened 2016-08-16
advertise: Jitter and request debouncing

abhinav:
This contains a series of fixes for our Hyperbahn advertise logic:

-   Don't start new `ad` requests if an existing request is ongoing.
-   Instead of retrying each request, let the next scheduled `ad` request take
    over if a request fails.
-   Add up to 5 seconds of jitter before the initial `ad` request. This will
    ensure that if lots of instances of the same service come up at the same
    time, they don't send `ad`s at the same time. Instead, they spread it out
    over 5 seconds.
-   Apply the same jitter to the interval for all consecutive `ad` requests.
    So instead of sending a request every 3 minutes, we'll send a request
    every 3 minutes plus up to 5 seconds.
-   If the initial `ad` request failed, instead of propagating a timeout
    error, propagate the real error.

Resolves https://github.com/uber/tchannel-python/issues/431

CC @blampe @breerly @kriskowal

### DIFF
--- a/tchannel/tornado/hyperbahn.py
+++ b/tchannel/tornado/hyperbahn.py
@@ -20,90 +20,140 @@
 
 from __future__ import absolute_import
 
+import sys
 import json
 import logging
 import random
-import time
 
 import tornado.gen
 import tornado.ioloop
 
-from ..errors import TimeoutError
 from .response import StatusCode
 
-EXPO_BASE = 1.4  # try this first
-MAX_DELAY = 10  # sec
-MAX_ATTEMPT = 7  # pow(1.4, 8) > 10
 DELAY = 3 * 60 * 1000  # ms delay time for successful advertise
-FIRST_ADVERTISE_TIME = 30  # sec
+PER_ATTEMPT_TIMEOUT = 2.0  # seconds
+FIRST_ADVERTISE_TIME = PER_ATTEMPT_TIMEOUT
+DEFAULT_INTERVAL_MAX_JITTER_SECS = 5
 
 log = logging.getLogger('tchannel')
 
 
-def _prepare_next_ad(attempt_counter):
-    delay_time = random.uniform(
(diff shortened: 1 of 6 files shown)


tchannel/tornado/hyperbahn.py
```diff
@@ -20,90 +20,140 @@
...
+
+        delay += random.uniform(0, self.interval_max_jitter_secs)
+        self._next_ad = self.io_loop.call_later(delay, self._ad,
+                                                response_future)
+
+    @tornado.gen.coroutine
+    def _ad(self, response_future=None):
+        self._next_ad = None
+        try:
+            response = yield self.tchannel.request(service='hyperbahn').send(
```
blampe: I suppose it would be too invasive to use the newer API here?

tchannel/tornado/hyperbahn.py
```diff
@@ -20,90 +20,140 @@
...
+                response_future.set_exc_info(sys.exc_info())
+        else:
+            if response.code != StatusCode.ok:
+                log.info('Failed to register with Hyperbahn: %s', response)
+            else:
+                log.info('Successfully registered with Hyperbahn')
+            if response_future is not None:
+                response_future.set_result(response)
+        finally:
+            self.io_loop.spawn_callback(self._schedule_ad)
```
blampe: Can't we call `self._schedule_ad()` directly here?

tchannel/tornado/hyperbahn.py
abhinav: Yeah, I don't know if that'll behave exactly how we want.

tchannel/tornado/hyperbahn.py
abhinav: Yeah, but this is slightly cheaper since it doesn't associate the stack context with the call. Made sense given that this is a background loop.

blampe:
LGTM

PR #436, opened 2016-08-19
Ignore Tracing field with empty/zero TraceID

yurishkuro:
Found this issue when testing with Java crossdock tests, because Java does not populate the Tracing field in the frame.

@abhinav


abhinav:
👍

yurishkuro:
@abhinav could you release a patch?

PR #439, opened 2016-08-23
Do not publish `crossdock` module, it's for tests only

yurishkuro:
I ran into similar issue with jaeger-client-python, but forgot to apply it here. The published egg should not contain `crossdock` module.

@abhinav


breerly:
Aaah, good catch! :shipit:

PR #441, opened 2016-08-26
Turn tracing on by default, but with Noop tracer

### LINKED ISSUE #440: Should tracing be ON by default?
Currently in the constructor the default is `trace=None` which ends up turning off the sampling of the spans to 0. 

Should this perhaps be `trace=True` by default?  If the `opentracing.tracer` is not initialized, then `trace=True` still results in no tracing, so it feels like an unnecessary switch, only for very special cases like Hyperbahn requests.

@abhinav @blampe @thepaulm

yurishkuro:
Fixes #440

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.28.2 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Change default setting for tracing to be enabled.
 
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
@blampe @abhinav

abhinav:
sgtm

abhinav:
Please update the changelog

PR #443, opened 2016-09-09
Validate that the headers object is actually a dict

yurishkuro:
When used with Raw encoding, the headers parameter contains
an unparsed string, and it was causing exceptions.

cc @abhinav


abhinav:
👍 
Thanks!

PR #444, opened 2016-09-12
Backport raw headers fix to 0.28

abhinav:
This backports Yuri's change from #443 to the 0.28 release.

CC @yurishkuro @breerly @blampe

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changes by Version
 ==================
 
+0.28.2 (unreleased)
+-------------------
+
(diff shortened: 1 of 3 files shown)


yurishkuro:
:shipit: 

thanks!

breerly:
:shipit:

breerly:
Curious - why backport this fix at all? Do we have customers that can't upgrade to the latest minor?

abhinav:
We haven't released 0.29 yet. We will probably do that soon but backporting it allows us to not worry about whether people have upgraded. Also, at least one customer wants to make use of the new API being introduced in 0.29, which means they'll probably need a little time to upgrade.

PR #445, opened 2016-09-19
Pass span.kind tag when calling start_span(), not after the span was started

yurishkuro:
Zipkin convention is to share span ID between client and server spans. Given OpenTracing API, that's only possible if `span.kind` tag is set when creating the span via `start_span()` method, not after.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changes by Version
 0.29.1 (unreleased)
 -------------------
 
-- Pass span.kind tag when calling start_span(), not after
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -4,7 +4,7 @@ Changes by Version
 0.29.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Pass span.kind tag when calling start_span(), not after
```
abhinav: nit: Use a full sentence here.

abhinav:
LGTM

PR #448, opened 2016-10-05
Relax opentracing upper bound to next major

yurishkuro:
@abhinav @blampe  can we release this asap?  It's breaking pip-compile for a number of services after recent opentracing upgrade.


abhinav:
lgtm

abhinav:
can you update the changelog before merging?

PR #449, opened 2016-10-05
peer selection: Never choose ephemeral peers

abhinav:
This changes peer selection to never choose ephemeral peers to send requests,
even if that is the only remaining peer.

Additionally, it loosens up the restrictions for what counts as an ephemeral
peer. Either the host can be `0.0.0.0` or the port can be `0` for a peer to be
an ephemeral. This is because Python TChannels that are not listening
advertise themselves as `127.0.0.1:0` (🙁).

---

Note to reviewers: Once this change is approved, I will backport it as a patch
release into older releases.

CC @blampe @breerly @prashantv

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ Changes by Version
 0.30.1 (unreleased)
 -------------------
 
-- Relax opentracing upper bound to next major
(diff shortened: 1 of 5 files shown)


blampe:
LGTM

PR #451, opened 2016-10-12
Propagate message send errors to the caller

abhinav:
This propagates message serialization and send errors up to the caller so that
we don't see cryptic `TimeoutError: None` exceptions instead.

Before:

```
    [..]
  File "[..]/tchannel-python/tchannel/tornado/peer.py", line 440, in _send
    response = yield response_future
  File "[..]/tchannel-python/env/lib/python2.7/site-packages/tornado-4.4.2-py2.7-macosx-10.11-x86_64.egg/tornado/gen.py", line 1015, in run
    value = future.result()
  File "[..]/tchannel-python/env/lib/python2.7/site-packages/tornado-4.4.2-py2.7-macosx-10.11-x86_64.egg/tornado/concurrent.py", line 237, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
tchannel.errors.TimeoutError: None``
```

After:

```
Traceback (most recent call last):
    [..]
  File "[..]/tchannel-python/tchannel/rw.py", line 574, in length_no_args
    size += rw.length(value)
  File "[..]/tchannel-python/tchannel/rw.py", line 617, in length
    size += self._pair.length(pair)
  File "[..]/tchannel-python/tchannel/rw.py", line 478, in length
    size += link.length(item)
  File "[..]/tchannel-python/tchannel/rw.py", line 446, in length
    s = s.encode('utf-8')
AttributeError: 'NoneType' object has no attribute 'encode'
```

@blampe @breerly @prashantv @willhug

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,9 @@ Changes by Version
 0.30.2 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Propagate message sending errors up to the caller. This should greatly reduce
+  the number of ``TimeoutError: None`` issues seen by users and show the root
+  cause instead.
 
 
 0.30.1 (2016-10-05)
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -301,12 +301,9 @@ class TornadoConnection(object):
             message_factory = self.response_message_factory
 
         fragments = message_factory.fragment(message)
-
-        for fragment in fragments:
-            future = self.writer.put(fragment)
-
-        # We're done writing the message once our last future resolves.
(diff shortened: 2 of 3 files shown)


prashantv:
Nice! lgtm

willhug:
Tests?

breerly:
Let's make this `1.0` please.

blampe:
Does this fail-fast in the case of a `None` name?

abhinav:
@willhug Added a test.

@blampe Yes, it fails with the specific failure in case of None. The test does
that. I'll make a separate change for #452.

PR #455, opened 2016-10-29
misc: fix minor connection.py issues

willhug:
1) InvalidMessageError no longer exists
2) Calling directly into the event_emitter (that doesn't exist)
3) Using logging.exception instead of log.exception


willhug:
@abhinav @breerly @kriskowal

PR #456, opened 2016-11-01
connection: Fixed fragmentation ordering regression

abhinav:
This fixes a regression in the behavior of how we handle large messages.
The correct behavior is to multiplex these frames among smaller requests
so that a large request doesn't block all other interaction until it is
done writing. Instead, we were queuing them all back to back. This was
also taking up a lot of memory because all fragments of a large message
would be held in memory until they were all written. With this change,
we won't get to the next fragment until the current fragment has been
serialized and written to the wire (`message_factory.fragment` is a
generator).

This also allows us to revert the revert of 0.30.2 since this takes care
of that memory leak as well. So we should see fewer obscure TimeoutError
exceptions.

CC @blampe @willhug

### DIFF
--- a/tchannel/tchannel.py
+++ b/tchannel/tchannel.py
@@ -33,7 +33,7 @@ from . import schemes
 from . import transport
 from . import retry
 from . import tracing
-from .errors import AlreadyListeningError
+from .errors import AlreadyListeningError, ServiceNameIsRequiredError
 from .glossary import DEFAULT_TIMEOUT
 from .health import health
 from .health import Meta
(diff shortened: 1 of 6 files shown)


tchannel/tchannel.py
```diff
@@ -101,6 +101,7 @@ def __init__(self, name, hostport=None, process_name=None,
             not provided an ephemeral port will be used. When advertising on
             Hyperbahn you callers do not need to know your port.
         """
+        assert name, 'service name cannot be empty or None'
```
blampe: Better error please, assertions belong in tests.

tests/test_tchannel.py
```diff
@@ -574,3 +574,14 @@ def healthy(request):
         thrift_module.Service.healthy(), caller_name='bar',
     )
     assert res.body is True
+
+
+@pytest.mark.parametrize("name", [
+    None,
+    "",
+])
+def test_service_name_is_required(name, io_loop):
```
willhug: don't need the ioloop right?

tests/tornado/test_connection.py
```diff
@@ -29,6 +29,7 @@
 
 from tchannel import TChannel
 from tchannel import messages
+from tchannel.tornado.request import Request
```
willhug: ordered imports?

PR #462, opened 2016-11-08
connection: Fix assert loop running errors

abhinav:
It seems that there are cases where a handshake is performed and before
the next call is finished, the connection disconnects. Before 0.30.4,
this would cause a StreamClosedError that would be retried, but now that
we're actually updating the `loop_running` variable, it causes an
assertion error.

The only reason we have `loop_running` is to track whether a handshake
was performed. With this change, we track it explicitly and don't unset
it once the handshake is performed. This should address the assertion
error people were seeing.

One of the asserts was removed because it has a `while not self.closed`
loop immediately afterwards, which would cause the loop to end.

@blampe @willhug

### DIFF
--- a/tchannel/tornado/connection.py
+++ b/tchannel/tornado/connection.py
@@ -391,6 +391,7 @@ class TornadoConnection(object):
                 "Expected handshake response, got %s" % repr(init_res)
             )
         self._extract_handshake_headers(init_res)
+        self._handshake_performed = True
 
         # The receive loop is started only after the handshake has been
         # completed.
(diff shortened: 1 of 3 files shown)


tchannel/tornado/connection.py
```diff
@@ -444,6 +442,7 @@ def _extract_handshake_headers(self, message):
         self.remote_host_port = int(self.remote_host_port)
         self.remote_process_name = message.process_name
         self.requested_version = message.version
+        self._handshake_performed = True
```
willhug: this isn't a handshake header.  We should set this explicitly in initiate_handshake and expect_handshake

abhinav:
released in 0.30.5

PR #463, opened 2016-11-14
handshake timeouts: retry with a different peer

abhinav:
If an outgoing handshake times out, rather than bubbling that timeout error
all the way to the caller, we should retry with a different peer.

@willhug @breerly

### DIFF
--- a/tests/test_tchannel.py
+++ b/tests/test_tchannel.py
@@ -40,10 +40,10 @@ from tornado.tcpserver import TCPServer
 
 from tchannel.tornado.stream import InMemStream
 from tchannel import TChannel, Request, Response, schemes, errors, thrift
-from tchannel.tornado import connection
 from tchannel.errors import AlreadyListeningError, TimeoutError
 from tchannel.event import EventHook
 from tchannel.response import TransportHeaders
+from tchannel.tornado import connection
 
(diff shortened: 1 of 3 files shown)


tests/test_tchannel.py
```diff
@@ -23,20 +23,24 @@
...
+import tornado
+
+from mock import MagicMock, patch, ANY
 from tornado import gen
+from tornado.netutil import bind_sockets
+from tornado.tcpserver import TCPServer
 
 from tchannel.tornado.stream import InMemStream
 from tchannel import TChannel, Request, Response, schemes, errors, thrift
+from tchannel.tornado import connection
```
willhug: order of imports

tests/test_tchannel.py
```diff
@@ -582,3 +586,35 @@ def test_service_name_is_required(name):
...
+    with patch.object(connection, 'DEFAULT_INIT_TIMEOUT_SECS', 0.1):
+        with pytest.raises(errors.NoAvailablePeerError):
+            yield tchannel.raw(service='server', endpoint='endpoint')
+
+
+@pytest.yield_fixture
+def timeout_server():
+
+    class HandshakeTimeoutServer(TCPServer):
+
```
willhug: maybe gratuitous spacing here

Issue #464, opened 2016-11-30
Calls made to specific host:port should not add to the peer list


prashantv:
If a call is made with `hostport=...`, then the value is added as a new peer to the peer list. If the user has a shared peer list, but explicitly wants to talk to some other service directly, then the pollution means the direct service host:port is now part of the shared peer lists, and any call could use this host:port.

Comment from @abhinav:
```
The fix for this is not too complicated. We need to change the peer
selection code to not add the hostport to the list of peers if it was
specified explicitly.

Right now, choose does get(hostport) when the hostport is
specified explicitly. get calls add if the hostport is not a known
peer.

https://github.com/uber/tchannel-python/blob/master/tchannel/tornado/peer.py#L706
https://github.com/uber/tchannel-python/blob/master/tchannel/tornado/peer.py#L637

Instead we will have to change it to build a Peer but not add it to
the list when hostport is specified.
```

PR #465, opened 2016-12-09
Don't add explicit hosts to the shared peer list, refs #464

blampe:
@willhug

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -702,8 +702,18 @@ class PeerGroup(object):
         """
 
         blacklist = blacklist or set()
+        # We we already have the connection in our peer list, use it. Else
+        # don't add it to our peer list.
+        # Refs https://github.com/uber/tchannel-python/issues/464
         if hostport:
-            return self.get(hostport)
+            if hostport in self.hosts:
+                return self.get(hostport)
+            else:
+                return self.peer_class(
+                    tchannel=self.tchannel,
+                    hostport=hostport,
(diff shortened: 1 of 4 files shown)


tchannel/tornado/peer.py
```diff
@@ -702,8 +702,18 @@ def choose(self, hostport=None, blacklist=None):
...
+        # Refs https://github.com/uber/tchannel-python/issues/464
         if hostport:
-            return self.get(hostport)
+            if hostport in self.hosts:
+                return self.get(hostport)
+            else:
+                return self.peer_class(
+                    tchannel=self.tchannel,
+                    hostport=hostport,
+                    on_conn_change=self._update_heap,
```
willhug: this peer will get added to the heap when it's updated.  We can set on_conn_change to None and it will get ignored

tchannel/tornado/peer.py
```diff
@@ -702,8 +702,18 @@ def choose(self, hostport=None, blacklist=None):
         """
 
         blacklist = blacklist or set()
+        # We we already have the connection in our peer list, use it. Else
+        # don't add it to our peer list.
+        # Refs https://github.com/uber/tchannel-python/issues/464
         if hostport:
-            return self.get(hostport)
+            if hostport in self.hosts:
```
willhug: we might want to just use
```
if hostport in self._peers:
```
That way (I assume) we're not converting the _peers dict into a list of keys on every choose.

willhug:
gonna commandeer

willhug:
creating a new pr

PR #466, opened 2016-12-09
Allow peerGroups to maintain one-off hostports without adding them to heap selection

willhug:
Summary: This PR modifies the peerGroup logic so that if a request is
sent directly to a Peer (through a hostport passed through the call
stack) instead of adding that peer to the peer heap, we will create it,
and only keep track of it in the `peers` map (not the peer heap).  This
means we will still maintain the peer, but we won't select it for the
regular hyperbahn instance.

Task: #464

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -592,18 +592,6 @@ class PeerGroup(object):
             self._peers = {}
             self._resetting = False
 
-    def get(self, hostport):
-        """Get a Peer for the given destination.
-
-        A new Peer is added and returned if one does not already exist for the
-        given host-port. Otherwise, the existing Peer is returned.
-        """
-        assert hostport, "hostport is required"
-        if hostport not in self._peers:
-            self.add(hostport)
-
-        return self._peers[hostport]
-
     def lookup(self, hostport):
         """Look up a Peer for the given host and port.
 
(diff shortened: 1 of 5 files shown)


tchannel/tornado/peer.py
```diff
@@ -621,30 +609,63 @@ def remove(self, hostport):
...
 
-        if isinstance(peer, basestring):
-            # Assume strings are host-ports
-            peer = self.peer_class(
-                tchannel=self.tchannel,
-                hostport=peer,
-                on_conn_change=self._update_heap,
+        if not isinstance(peer, basestring):
+            # Assume non-strings are Peers
+            assert peer.hostport not in self._peers, (
```
willhug: this doesn't account for the case where someone is trying to add a peer object for a peer that already exists, but we don't really know what to do in that case anyway... which peer would we take?

tchannel/tornado/peer.py
```diff
@@ -681,12 +730,6 @@ def request(self, service, hostport=None, **kwargs):
             hostport=hostport,
             **kwargs)
 
-    def _connected_peers(self, hostports):
```
willhug: dead func, not called

blampe:
Clearly I underestimated the fix :) Thank you for taking this on!

This looks OK to me and tests pass so I'm good with it. We won't cut this today and Abhinav can give you some eyes on Monday.

tchannel/tornado/peer.py
blampe: If we take the one that already exists would that re-use the existing connection? I think we want that.

tests/tornado/test_peer.py
```diff
@@ -324,8 +324,9 @@ def test_choose_with_target_hostport(hostports):
     target = '1.0.0.1:9000'
     for _ in hostports:
         peer = peer_group.choose(hostport=target)
+        assert target in peer_group.hosts
         assert target == peer.hostport
-        assert peer_group.peer_heap.size() == n
+        assert peer_group.peer_heap.size() == n - 1
```
prashantv: `n` is set to `len(hostports) + 1`, but the only use of `n` does `n - 1`, maybe we should do: `n = len(hostports)`?

tests/tornado/test_peer.py
```diff
@@ -324,8 +324,9 @@ def test_choose_with_target_hostport(hostports):
     target = '1.0.0.1:9000'
     for _ in hostports:
```
prashantv: don't really understand why there's a `for` loop here

tests/tornado/test_peer.py
```diff
@@ -346,3 +347,27 @@ def hello(request):
         'choose() MUST NOT select the ephemeral peer even if that is the only'
         'available peer'
     )
+
+
+def test_upgrade_target_host_to_direct(hostports):
```
prashantv: same comments as above test --don't think we need the second `for` loop

tchannel/tornado/peer.py
```diff
@@ -621,30 +609,63 @@ def remove(self, hostport):
...
-        if peer:
+        peer_in_heap = peer and peer.index != -1
+        if peer_in_heap:
             self.peer_heap.remove_peer(peer)
         return peer
 
+    def get(self, hostport):
+        """Get a Peer for the given destination.
+
+        A new Peer is added and returned if one does not already exist for the
```
prashantv: call out that `get` adds a peer to the peer heap

tchannel/tornado/peer.py
```diff
@@ -621,30 +609,63 @@ def remove(self, hostport):
...
+            self._peers[peer.hostport] = peer
+            self._add_to_heap(peer)
+            return
 
-        assert peer.hostport not in self._peers, (
-            "%s already has a peer" % peer.hostport
+        hostport = peer
+        if hostport in self._peers:
+            # upgrade a direct peer to a heap peer
+            self._add_to_heap(self._peers[hostport])
```
prashantv: do you need to verify whether the specified peer is already in the heap? or does the `add_to_heap` handle this?

tchannel/tornado/peer.py
```diff
@@ -656,6 +677,34 @@ def _update_heap(self, peer):
...
+
+        **NOTE** new peers will not be added to the peer heap.
+        """
+        assert hostport, "hostport is required"
+        if hostport not in self._peers:
+            self.add_direct(hostport)
+
+        return self._peers[hostport]
+
+    def add_direct(self, hostport):
```
prashantv: Should all the new methods be underscore prefixed since I don't think they should be part of the public API surface

tchannel/tornado/peer.py
```diff
@@ -656,6 +677,34 @@ def _update_heap(self, peer):
...
+        """
+        Add a peer directly from a hostport, do NOT add it to the peer heap
+        """
+        peer = self.peer_class(
+            tchannel=self.tchannel,
+            hostport=hostport,
+        )
+
```
prashantv: do you need a `assert hosrtport, "hostport is required"` here?

willhug: shooooooooouldn't need it, but paranoia, yay

hmm, this function is pretty small without that, maybe we can just inline it...

tchannel/tornado/peer.py
```diff
@@ -656,6 +677,34 @@ def _update_heap(self, peer):
...
+        """Get a Peer for the given destination for a request.
+
+        A new Peer is added and returned if one does not already exist for the
+        given host-port. Otherwise, the existing Peer is returned.
+
+        **NOTE** new peers will not be added to the peer heap.
+        """
+        assert hostport, "hostport is required"
+        if hostport not in self._peers:
+            self.add_direct(hostport)
```
abhinav: `add_direct` isn't used anywhere else. Just inline it here.

tchannel/tornado/peer.py
```diff
@@ -656,6 +677,34 @@ def _update_heap(self, peer):
         peer.rank = rank
         self.peer_heap.update_peer(peer)
 
+    def get_for_request(self, hostport):
```
abhinav: Consider calling this `get_isolated` or something similar and make it private while you're at it.

tchannel/tornado/peer.py
willhug: it does now :D

PR #467, opened 2016-12-10
Add str functions to peer and PeerClientOperation for better debugability from exc_info

willhug:


prashantv:
lgtm, really wanted this while debugging with larry yesterday.

PR #474, opened 2017-10-09
Yield hooks, add a new hook to allow modification of application headers

lucasmarshall:
It is not currently possible to make hooks asynchronous, since you can't supply a coroutine as a hook method because they wouldn't be yielded properly. This fixes that.

### DIFF
--- a/tchannel/event.py
+++ b/tchannel/event.py
@@ -36,6 +36,7 @@ EventType = enum(
     'EventType',
     before_send_request=0x00,
     after_send_request=0x01,
+    before_serialize_request_headers=0x02,
     before_send_response=0x10,
     after_send_response=0x11,
     before_receive_request=0x20,
@@ -61,6 +62,12 @@ class EventHook(object):
                 ....
 
     """
+
+    def before_serialize_request_headers(self, headers):
+        """Called before an outgoing request is serialized.
+        Only to be used for modifying application headers.
+        """
+        pass
 
     def before_send_request(self, request):
         """Called before any part of a ``CALL_REQ`` message is sent."""
--- a/tchannel/event.py
+++ b/tchannel/event.py
@@ -154,7 +154,9 @@ class EventEmitter(object):
     def fire(self, event, *args, **kwargs):
         for hook in self.hooks[event]:
             try:
-                yield tornado.gen.maybe_future(hook(*args, **kwargs))
+                possible_future = hook(*args, **kwargs)
+                if tornado.concurrent.is_future(possible_future):
+                    yield possible_future
             except Exception:
                 log.error("error calling hook", exc_info=sys.exc_info())
 
--- a/tchannel/event.py
+++ b/tchannel/event.py
@@ -24,6 +24,7 @@ import sys
 import collections
 import functools
 import logging
+import tornado
 
 from .enum import enum
 
@@ -60,6 +61,7 @@ class EventHook(object):
                 ....
 
     """
+
     def before_send_request(self, request):
         """Called before any part of a ``CALL_REQ`` message is sent."""
         pass
@@ -141,10 +143,11 @@ class EventEmitter(object):
                 event_value = getattr(EventType, event_type)
                 self.register_hook(func, event_value)
 
+    @tornado.gen.coroutine
     def fire(self, event, *args, **kwargs):
         for hook in self.hooks[event]:
             try:
-                hook(*args, **kwargs)
+                yield tornado.gen.maybe_future(hook(*args, **kwargs))
             except Exception:
                 log.error("error calling hook", exc_info=sys.exc_info())
 
--- a/tchannel/event.py
+++ b/tchannel/event.py
@@ -63,7 +63,7 @@ class EventHook(object):
 
     """
 
-    def before_serialize_request_headers(self, headers):
+    def before_serialize_request_headers(self, headers, service):
         """Called before an outgoing request is serialized.
         Only to be used for modifying application headers.
         """
--- a/tchannel/schemes/raw.py
+++ b/tchannel/schemes/raw.py
@@ -22,6 +22,7 @@ from __future__ import (
     absolute_import, division, print_function, unicode_literals
 )
 
+from ..event import EventType
 from . import RAW
 
 
@@ -102,6 +103,10 @@ class RawArgScheme(object):
 
         :rtype: Response
         """
+        yield self.tchannel.event_emitter.fire(
+            EventType.before_send_request_headers,
+            headers,
+        )
 
         return self._tchannel.call(
             scheme=self.NAME,
--- a/tchannel/schemes/raw.py
+++ b/tchannel/schemes/raw.py
@@ -109,6 +109,7 @@ class RawArgScheme(object):
         yield self._tchannel._dep_tchannel.event_emitter.fire(
             EventType.before_serialize_request_headers,
             headers,
+            service,
         )
 
         response = yield self._tchannel.call(
--- a/tchannel/schemes/raw.py
+++ b/tchannel/schemes/raw.py
@@ -22,6 +22,8 @@ from __future__ import (
     absolute_import, division, print_function, unicode_literals
 )
 
+from tornado import gen
+
 from ..event import EventType
 from . import RAW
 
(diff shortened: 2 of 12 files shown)


lucasmarshall:
@abhinav @breerly

breerly:
Hey @lucasmarshall - can I get some background into the motivation of this change? In practice, this is an unmaintained library - every feature we ship here means additional docs, releases, announcements, etc...

Just want to be sure this is a need :)

lucasmarshall:
@breerly currently galileo is using the `before_send_request` hook to do instrumentation of outbound requests, and that can sometimes result in network IO to wonka, so `before_send_request` really needs to be able to be able to run asynchronously to not hold up the IOLoop.

lucasmarshall:
As far as documentation, really the only thing that would need to be communicated is that your hooks can be (but don't need to be) implemented as coroutines.

lucasmarshall:
It seems that the travis CI job that runs pypy tests is broken, FYI.

tchannel/tornado/peer.py
```diff
@@ -430,7 +430,7 @@ def send(
                 )
         except Exception as e:
             # event: on_exception
-            self.tchannel.event_emitter.fire(
+            yield self.tchannel.event_emitter.fire(
                 EventType.on_exception, request, e,
             )
             raise
```
Groxx: this will not work - the exception will be lost after the yield.
you can save the `sys.exc_info()` before yielding and raise it after, but there are some subtleties with old vs new style exception-raising (i.e. `raise Exception, 'message'` vs `raise Exception('message')`, which means you can't just `raise exc[0], exc[1], exc[2]` in all cases.

lucasmarshall: Good catch. Will fix.

lucasmarshall:
@blampe

tchannel/event.py
```diff
@@ -60,6 +61,7 @@ def before_send_request(self, request):
                 ....
 
     """
+
```
willhug: why the space?

willhug:
Hmm, looks like we missed one [here](https://github.com/uber/tchannel-python/blob/master/tchannel/tornado/connection.py#L536)).  Unfortunately this is not in an async context so I'm not really sure what the proper way to "asyncify" it is.

One a side note, since we only really need this for one of the events (`before_send_request` I believe) does it make sense to add a new function to the event emitter that is async by default?  So we could add a an `async_fire`.  This would reduce the scope of the change, and isolate changes only to the places that need the new feature.  (In general we are very hesitant to making sweeping changes to this codebase).

willhug:
Also, we should add to the changelog :D

lucasmarshall:
@willhug Yeah, I intentionally missed that one. Not sure how it would make sense to do a yield there.

As for making an async_fire, I'm not sure that really would make sense. That would make it so that the programmer writing a hook would have to know too much about the internal tchannel implementation to know if they could implement an async hook or not. I think that consistency is important here.

I will update the changelog.

tchannel/event.py
lucasmarshall: I use autopep8. It added it automatically

tchannel/event.py
```diff
@@ -35,6 +36,7 @@
     'EventType',
     before_send_request=0x00,
     after_send_request=0x01,
+    before_serialize_request_headers=0x02,
```
willhug: wait, this should be "send", disregard other comments.

lucasmarshall:
@willhug `run_sync` won't work, unfortunately. What `run_sync` does is:
1. starts the IOLoop
1. runs the function
1. stops the IOLoop

Unfortunately, if you have an already running IOLoop step 1 throws an exception. So any tornado-based application would break. Actually, reading through @breerly's implementation of [threadloop](https://github.com/breerly/threadloop) (which tchannel uses for its "sync" implementation), it would break in non-tornado apps too as that just runs an IOLoop on a separate thread.

What we probably want to do is use `IOLoop.add_callback`, which will call the method on the next loop.

tchannel/tornado/connection.py
```diff
@@ -533,9 +533,11 @@ def send_error(self, error):
         error_message = build_raw_error_message(error)
         write_future = self.writer.put(error_message)
         write_future.add_done_callback(
-            lambda f: self.tchannel.event_emitter.fire(
-                EventType.after_send_error,
-                error,
+            lambda f: IOLoop.current().add_callback(
```
willhug: I think we might still have an issue here, it will still execute the coroutine as if it were a regular function (thus possibly creating a race condition), but it's proooooobably fine since all the after_send_error does is log some stats, and I think we have some guarantee that the code will be at least executed.  Please correct me if I'm wrong on that.
cc @abhinav, @Groxx

tchannel/event.py
Groxx: why send?

Groxx: ah, because other things call it with "send".  initially it was send, but I voted to change it to "serialize" since all comments around it / its purpose was for before-serialization.

unless there's a reason to prefer "send" in the event name I'll vote for the calls to be "serialize" too.

tchannel/event.py
```diff
@@ -141,10 +150,11 @@ def register_hook(self, hook, event_type=None):
                 event_value = getattr(EventType, event_type)
                 self.register_hook(func, event_value)
 
+    @tornado.gen.coroutine
     def fire(self, event, *args, **kwargs):
         for hook in self.hooks[event]:
             try:
-                hook(*args, **kwargs)
+                yield tornado.gen.maybe_future(hook(*args, **kwargs))
```
Groxx: depending on how non-yielded performance-oriented we are: there's an `is_future` function, it might be measurably faster to only yield when necessary, since this'll mean (for our service) 3 round-trips to the IOLoop for each event (we have 3 hooks).

`is_future` is what maybe_future uses fwiw, should be reliable.

lucasmarshall: Sounds reasonable. I'll update.

tchannel/schemes/raw.py
```diff
@@ -33,6 +36,7 @@ class RawArgScheme(object):
     def __init__(self, tchannel):
         self._tchannel = tchannel
 
+    @gen.coroutine
     def __call__(
```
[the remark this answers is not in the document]
lucasmarshall: It actually did, but only because self._tchannel.call() returns a future.

tchannel/tornado/connection.py
Groxx: pretty sure it won't work as-is: http://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.add_callback
should be something like `add_callback(self.tchannel.event_emitter.fire, EventType.after_send_error, error)`

but yeah, afaik this is perfectly safe.  even if you don't explicitly yield it / put it in the ioloop, it still *runs* with the same guarantees as everything else (minus the fact that you've given up ordering), it just complains in logs that you never retrieved the value / exception.

lucasmarshall: Actually the future does get resolved. `add_callback` causes the function to be called in the next loop, using `_run_callback`. `_run_callback` calls the function and does a `add_future` if the call results in a future (as this one will).

Ref: https://github.com/tornadoweb/tornado/blob/d68abdb4b68632f675f41ff817c370dbc8f1e21f/tornado/ioloop.py#L659

Groxx: Ah, derp, yeah - I just realized that's what he meant.  Thanks!

Groxx: As far as "send eventually"... given Writer's implementation (via Queue), it may not even be a behavioral change, aside from being slightly later (while still async).  I'm not familiar enough with the code to be sure about that though :|

Pragmatically, it doesn't look like you can / should expect to be able to change the error in the hook, so I don't think this makes a practical difference (especially since I think it's currently async too).

Seems fine.  Probably worth documenting if we think it does change the behavior, but anything it breaks seems like it'd have to be an un-supported use-case.

tchannel/event.py
willhug: yeah, sgtm

tchannel/tornado/connection.py
willhug: nice!

PR #475, opened 2017-10-11
Added Client Interceptors to tracing implementation

dray92:
Depends on merging [this one](https://github.com/uber-common/opentracing-python-instrumentation/pull/31)

Looking for comments on approach

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ certifi==2016.2.28        # via tornado
 contextlib2==0.5.1        # via opentracing-instrumentation
 crcmod==1.7
 futures==3.0.5            # via opentracing, opentracing-instrumentation
-opentracing-instrumentation==2.2.0
+opentracing-instrumentation==2.3.0
 opentracing==1.1.0        # via opentracing-instrumentation
 ply==3.8                  # via thriftrw
 singledispatch==3.4.0.3   # via tornado
--- a/setup.py
+++ b/setup.py
@@ -58,10 +58,10 @@ setup(
 
         # tracing deps
         'opentracing>=1.1,<2',
-        'opentracing_instrumentation>=2.3,<3'
+        'opentracing_instrumentation>=2.3,<3',
 
         # python 3 compat
-        'six'
+        'six',
     ],
     extras_require={
         'vcr': ['PyYAML', 'mock', 'wrapt'],
--- a/setup.py
+++ b/setup.py
@@ -58,10 +58,10 @@ setup(
 
         # tracing deps
         'opentracing>=1.1,<2',
-        'opentracing_instrumentation>=2,<3',
+        'opentracing_instrumentation>=2.3,<3'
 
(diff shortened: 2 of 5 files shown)


debosmit:
@yurishkuro

abhinav:
Hi. Can you explain in the PR description what this change is doing and why?
The TChannel Python library is in maintenance mode at this time and we're
actively avoiding adding any new features to it. There has to be a very strong
reason for new functionality to be added to it.

setup.py
```diff
@@ -58,10 +58,10 @@
 
         # tracing deps
         'opentracing>=1.1,<2',
-        'opentracing_instrumentation>=2,<3',
+        'opentracing_instrumentation>=2.3,<3'
```
willhug: we've removed the comma here.  The setup.py can't parse the dependencies here.  We should add back the commas to the end of each dependency.  This is causing the tests to fail (apart from the pypy tests which appear to be always failing).

setup.py
```diff
@@ -58,10 +58,10 @@
...
 
         # python 3 compat
-        'six',
+        'six'
```
willhug: ditto on the comma at the end of the line

requirements.txt
```diff
@@ -10,7 +10,7 @@ certifi==2016.2.28        # via tornado
 contextlib2==0.5.1        # via opentracing-instrumentation
 crcmod==1.7
 futures==3.0.5            # via opentracing, opentracing-instrumentation
-opentracing-instrumentation==2.2.0
```
willhug: was this changed via `make requirements.txt`?

We should make sure that works since I beleive that's the code that runs the setup.py stuff.

debosmit: I believe I did that -- running `make requirements.txt` again just to be sure

tchannel/schemes/thrift.py
```diff
@@ -133,7 +133,8 @@ def __call__(
 
         span, headers = self.tracer.start_span(
             service=request.service, endpoint=request.endpoint,
-            headers=headers, hostport=hostport, encoding='thrift'
+            headers=headers, hostport=hostport, encoding='thrift',
+            request=request
```
abhinav: Don't you need to update the other start_span sites?

tchannel/schemes/thrift.py
```diff
@@ -141,8 +143,15 @@ def __call__(
             headers,
             request.service,
         )
-        serializer = request.get_serializer()
 
+        # fire interceptors
+        for interceptor in ClientInterceptors.get_interceptors():
+            if isinstance(interceptor, TChannelOpenTracingClientInterceptor):
+                interceptor.process(span=span, request=request,
```
abhinav: As discussed offline, we should remove the request object from here.

lucasmarshall:
Can we please get a new version cut?

abhinav:
Released in 1.3.0

PR #479, opened 2018-05-30
peer: Don't send requests to incoming peers

### LINKED ISSUE #478: tchannel-python assumes any incoming connections with host:port in init headers can be used to send RPCs
When TChannel receives an incoming connection, it receives init headers which include a host:port that represents whether the other side is listening.

tchannel-python currently assumes that any connection which has a host:port is OK to send RPCs back on. It previously also sent RPCs to connections that were marked ephemeral, but [that was fixed](https://github.com/uber/tchannel-python/pull/449).

If any service connects to a tchannel-python service, RPCs may end up going back over that incoming connection, and typically the client receives a `BadRequest` error (because the remote host:port was not expecting those RPCs). These requests are not retried, and cause an immediate failure.

abhinav:
This alters peer selection to stop considering peers that send incoming
requests as valid candidates to send outgoing requests to. This was
previously partially fixed in #449 where we stopped considering
ephemeral peers, but non-ephemeral peers are still an issue.

The only peers we should consider for outgoing requests are those we
explicitly added to the peer heap.

Resolves #478

### DIFF
--- a/tchannel/tornado/peer.py
+++ b/tchannel/tornado/peer.py
@@ -69,6 +69,7 @@ class Peer(object):
         'chosen_count',
         'on_conn_change',
         'connections',
+        'is_incoming',
 
         '_connecting',
         '_on_conn_change_cb',
@@ -104,6 +105,9 @@ class Peer(object):
         #: are added to the left side of the deque and outgoing connections to
         #: the right side.
         self.connections = deque()
+
+        # Whether this peer has incoming connections only.
+        self.is_incoming = False
 
         # This contains a future to the TornadoConnection if we're already in
         # the process of making an outgoing connection to the peer. This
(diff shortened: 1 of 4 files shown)


tchannel/tornado/peer.py
```diff
@@ -727,6 +736,11 @@ def choose(self, hostport=None, blacklist=None):
...
 
-        return self.peer_heap.smallest_peer(
-            (lambda p: p.hostport not in blacklist and not p.is_ephemeral),
-        )
+        def is_valid_peer(p):
+            return not (
+                p.hostport in blacklist or
+                p.is_ephemeral or
+                p.is_incoming
+            )
```
akshayjshah: Nice, this is actually quite a bit clearer than the lambda.

tchannel/tornado/peer.py
```diff
@@ -727,6 +736,11 @@ def choose(self, hostport=None, blacklist=None):
         if hostport:
             return self._get_isolated(hostport)
 
-        return self.peer_heap.smallest_peer(
-            (lambda p: p.hostport not in blacklist and not p.is_ephemeral),
-        )
+        def is_valid_peer(p):
```
akshayjshah: Perhaps `should_send_requests`? Given the number of concerns here, "valid" is pretty vague.

PR #482, opened 2018-06-05
Ignore incoming peers only if they're unknown

### LINKED ISSUE #478: tchannel-python assumes any incoming connections with host:port in init headers can be used to send RPCs

abhinav:
The fix for #478 in f8fa6dc was broken. It ignored any peer that was
providing incoming connections only. This doesn't work if the peer
connecting to you is a known router being advertised to.

The correct fix, as suggested by @prashantv in the original issue is to
ignore incoming peers from the peer heap if they're not already known.
This reverts the change made to the peer selection logic in f8fa6dc
while keeping the test and adds the suggested fix instead. This also
adds a test that fails both, without this change and without f8fa6dc.

Resolves #478

### DIFF
--- a/tests/test_hyperbahn.py
+++ b/tests/test_hyperbahn.py
@@ -56,7 +56,7 @@ def client(io_loop):
 
 @pytest.yield_fixture
 def routers(server, io_loop):
-    routers = [TChannel('hyperbahn') for i in range(5)]
+    routers = [TChannel('hyperbahn') for i in xrange(5)]
 
     # Generates an ad function for routers[i].
     def ad_for(me):
--- a/tests/test_hyperbahn.py
+++ b/tests/test_hyperbahn.py
@@ -44,6 +44,9 @@ def server(io_loop):
 @pytest.yield_fixture
 def client(io_loop):
     client = TChannel('client')
+    # The client listens for incoming connections so that the server doesn't
+    # treat is as an ephemeral client. This is needed to exercise the p2p use
+    # case.
     client.listen()
     try:
         yield client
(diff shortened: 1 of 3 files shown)


abhinav:
CC @akshayjshah

tests/test_hyperbahn.py
```diff
@@ -0,0 +1,120 @@
...
+    try:
+        yield server
+    finally:
+        server.close()
+
+
+@pytest.yield_fixture
+def client(io_loop):
+    client = TChannel('client')
+    client.listen()
```
prashantv: add a comment about why this is important

tests/test_hyperbahn.py
```diff
@@ -0,0 +1,120 @@
...
+    try:
+        yield client
+    finally:
+        client.close()
+
+
+@pytest.yield_fixture
+def routers(server, io_loop):
+    routers = [TChannel('hyperbahn') for i in range(5)]
```
akshayjshah: Uh hello, `xrange` for mo' betta efficiency. Gotta count those nanos.

abhinav: wontfix. this breaks python 3 compatibility.

PR #483, opened 2018-08-03
Update OpenTracing instrumentation package

isaachier:


abhinav:
Sorry, I don't see how this will fix an issue. The original version bound is
broad enough that 2.4 will be picked up by the tooling. You're likely getting
the older version from a different dependency. I think you can run `pip` with
three `-v`s to get super-verbose output. That may tell you why it's picking up
the old version.

isaachier:
Fair enough. Not really sure what was going on in my build but it seemed to grep for identical versions of instrumentation and fail there. Will update soon. Thanks.

abhinav:
https://github.com/naiquevin/pipdeptree may prove useful in debugging this

isaachier:
OK seems I picked the wrong package. Sorry for the confusion. Thanks again!

isaachier:
Awesome. Actually turned out to be a tox issue. Pinned version of OT instrumentation was wrong there. Sorry for the hassle.

PR #486, opened 2019-07-12
Repairing the build

westover:


westover:
Ok so it looks like the default build is broken now for coverage because the default python on Travis is python 3.6

PR #489, opened 2019-07-13
Python3.4 compatibility

westover:


tchannel/testing/vcr/server.py
```diff
@@ -154,9 +157,11 @@ def send(self, request):
         try:
             response = yield response_future
         except TChannelError as e:
+            import traceback
```
abhinav: already imported above?

westover: Removed good catch

westover:
@willsalz fyi

PR #490, opened 2019-10-01
Update tracing to OpenTracing api 2

westover:


tchannel/tracing.py
```diff
@@ -97,7 +97,11 @@ def span_in_context(self, span):
         :param span: an OpenTracing Span
         :return: ``StackContext``-based context manager
         """
-        return opentracing_instrumentation.span_in_stack_context(span)
+        if isinstance(opentracing.tracer.scope_manager,
+                      opentracing_instrumentation.request_context.TornadoScopeManager):  # noqa
+            return opentracing_instrumentation.span_in_stack_context(span)
+
+        return opentracing_instrumentation.span_in_context(span)
```
abhinav: there doesn't appear to be a difference between the if and else cases. Is this a leftover from before?

westover: span_in_stack_context vs span_in_context

tchannel/tracing.py
```diff
@@ -97,7 +97,11 @@ def span_in_context(self, span):
         :param span: an OpenTracing Span
         :return: ``StackContext``-based context manager
         """
-        return opentracing_instrumentation.span_in_stack_context(span)
+        if isinstance(opentracing.tracer.scope_manager,
+                      opentracing_instrumentation.request_context.TornadoScopeManager):  # noqa
```
abhinav: suspect you can make this pep8 friendly with

```
if isinstance(
    opentracing.tracer.scope_manager,
    opentracing_instrumentation.request_context.TornadoScopeManager,
):
    ...
```

westover: Nope still too long :)

setup.py
```diff
@@ -57,8 +57,8 @@
         'threadloop>=1,<2',
 
         # tracing deps
-        'opentracing>=1.1,<2',
-        'opentracing_instrumentation>=2.3,<3',
+        'opentracing>2',
+        'opentracing_instrumentation>3',
```
willsalz: so…this now conflicts with jaeger-client-python, which requires opentracing_instrumentation<3

westover: jaeger-client-python is internal to Uber so lets take this convo to uChat :)

PR #491, opened 2019-10-18
Updating to py37 compatibility

westover:
Explicitly support py36 and py37


westover:
PY35 is explicitly unsupported at this moment

PR #496, opened 2022-02-04
Add version conditioning on futures and context

westover:


abhinav (approved):
appears to be [environment markers syntax](https://www.python.org/dev/peps/pep-0508/#environment-markers)

